PR #14, opened 2022-11-04
Rename indy-credx to anoncreds

dkulic:


swcurran:
Should we have a requirement as part of evolving this library to make parallel changes in one or more existing Aries Frameworks (or would it be indy-shared-rs?) so that we have relevant tests that demonstrate the change here works in Aries implementations?  We need to do that at some point and I'm thinking we get that practice in place early, vs. getting lots done here and adding the tests that matter to implementers later.

TimoGlastra (approved):
Nice 👍 

I see tests are not running for this PR, would be good to enable these somehow. I assume we can just take that over from the indy-shared-rs repo?

TimoGlastra:
Ah it's probably because this is pointed at a custom branch. I'm going to set the target to main

TimoGlastra:
Think we need to push some code to run the tests...

dkulic:
To push something in this PR? Not sure I understood?

TimoGlastra:
Ah that was indeed a bit vague. The PR was first opened against a custom branch in this repo which didn't trigger CI. I changed the PR to merge into main, but it didn't trigger the CI on the main branch, so I think we need to push some  reopen the PR to trigger the CI on this PR

dkulic:
Sorry for the delay, I have pushed some small change, looks like it didn't run the tests?

TimoGlastra:
Not sure why it isn't running the CI should run on PRs to main: 

```
on:
  push:
    branches: [main]
  pull_request:
    branches: [main]
```

https://github.com/hyperledger/anoncreds-rs/blob/main/.github/workflows/rusttest.yml

dkulic:
Do you want me to merge this one, or to keep experimenting? I run tests locally.
Unfortunately, I do not have much experience with setting this up.

TimoGlastra:
Let's merge for now, and pick it up as a separate task

PR #18, opened 2022-11-11
Moved indy-data-types into anoncreds module data_types

dkulic:
Moved indy-data-types inside of anoncreds crate.
In this process I removed "serde", "cl" and "cl_native" features (because they are mandatory for the anoncreds code).
Other features are moved.


dkulic:
There is an issue, that compiling fails if started within anoncreds dir. From top level works fine. Looking into this.

dkulic:
> There is an issue, that compiling fails if started within anoncreds dir. From top level works fine. Looking into this.

Fixed

anoncreds/src/data_types/anoncreds/cred_def.rs
```diff
@@ -26,18 +25,17 @@ impl SignatureType {
     }
 }
 
-#[derive(Debug)]
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug, Serialize, Deserialize)]
 pub struct CredentialDefinitionData {
     pub primary: ursa_cl!(CredentialPrimaryPublicKey),
-    #[cfg_attr(feature = "serde", serde(skip_serializing_if = "Option::is_none"))]
+    #[serde(skip_serializing_if = "Option::is_none")]
```
blu3beri, line 31: Is Serde not going to be a feature anymore?

anoncreds/src/data_types/anoncreds/macros.rs
```diff
@@ -0,0 +1,5 @@
+macro_rules! ursa_cl {
```
blu3beri, line 1: Very unfamiliar with this specific macro, but why can't we directly just import the package?

anoncreds/src/data_types/anoncreds/cred_def.rs
blu3beri: Ah I see that it came from another macro which was based on a config on whether cl/cl_native was enabled. 

Right now, this macro can be removed and it can be imported directly as we always pull in cl/cl_native right now.

anoncreds/Cargo.toml
```diff
@@ -19,29 +19,29 @@ crate-type = ["staticlib", "rlib", "cdylib"]
...
-sha2 = "0.9"
+serde_json = { version = "1.0", features = ["raw_value"]}
+sha2 = "0.10"
 tempfile = "3.1.0"
 thiserror = "1.0"
-zeroize = { version = "1.3", optional = true }
+ursa = { version = "=0.3.6", default-features = false, features = ["cl_native", "serde"] }
+zeroize = { version = "1.3", optional = true, features = ["zeroize_derive"] }
+hex = { version = "0.4", optional = true }
+openssl = { version = "0.10", optional = true }
```
blu3beri, line 40: Where are openssl and hex used?

blu3beri (changes requested):

anoncreds/src/data_types/anoncreds/macros.rs
blu3beri: Ah I see that it came from another macro which was based on a config on whether cl/cl_native was enabled.

Right now, this macro can be removed and it can be imported directly as we always pull in cl/cl_native right now.

dkulic: Good catch.

anoncreds/src/data_types/anoncreds/cred_def.rs
dkulic: Yeah, it was enabled with cl_native feature, but these features are required for anoncreds code, so I removed them (no need for conditional code).

anoncreds/src/data_types/anoncreds/macros.rs
dkulic: Removed

anoncreds/Cargo.toml
dkulic: Only in copied features vendored and merkle_tree.

anoncreds/src/lib.rs
```diff
@@ -7,8 +7,12 @@ extern crate log;
 #[macro_use]
 extern crate serde;
 
+#[cfg(test)]
+#[macro_use]
+extern crate serde_json;
+
 #[doc(hidden)]
-pub use indy_data_types::ursa;
+pub use ursa;
```
blu3beri, line 15: I don't think we have to also supply hit here, but we can leave it for now.

blu3beri (approved):

anoncreds/Cargo.toml
blu3beri: Yeah would be good if we can remove the openssl dependency (we still pull it in via `Ursa` though) as it can cause quite some issues when building for different platforms.

dkulic:
Do you want to first merge your PR, there will be some conflicts I think?

blu3beri:
> @blu3beri Do you want to first merge your PR, there will be some conflicts I think?

I have no preference, I can solve the merge conflicts as well :). I think its just the features and the `lib.rs` in `anoncreds/`

dkulic:
Your PR was first there, so go ahead :)

PR #21, opened 2022-11-18
Removed to_unqualified() method from Credential Definition data type

dkulic:
Looks like the datatype is already in line with the spec

### DIFF
--- a/anoncreds/src/data_types/anoncreds/cred_def.rs
+++ b/anoncreds/src/data_types/anoncreds/cred_def.rs
@@ -45,20 +45,6 @@ impl CredentialDefinition {
             CredentialDefinition::CredentialDefinitionV1(c) => &c.id,
         }
     }
-
-    pub fn to_unqualified(self) -> CredentialDefinition {
-        match self {
-            CredentialDefinition::CredentialDefinitionV1(cred_def) => {
-                CredentialDefinition::CredentialDefinitionV1(CredentialDefinitionV1 {
-                    id: cred_def.id.to_unqualified(),
-                    schema_id: cred_def.schema_id.to_unqualified(),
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            lib: libindy_credx.so
+            lib: anoncreds.so
```
andrewwhitehead, line 21: This should likely be libanoncreds.so, and libanoncreds.dylib below, just to match platform standards.

dkulic: Correct. Not sure how I wrote it this way :)

blu3beri:
I see that the Github bot is throwing quite some warnings about unused code. Is this something we will use in the future or can we remove it already?

dkulic:
I was not sure should I remove it.
I think for that macro, there is a low chance to be used (and if yes, it is easy to write it again).
Regarding the `RequestedPredicate`, my guess is it will be needed, but again not 100% sure.
And I have an easy fix for the last warning (but avoided changing the code in other package).

Let me remove/change what I think is mostly safe.

blu3beri (approved):
LGTM.

TimoGlastra:
> Looks like the datatype is already in line with the spec

@dkulic I think the schema_id property isn't in line yet. The spec uses snake_case for it, while the anoncreds implementation currently transforms it to camelCase

TimoGlastra:
> @dkulic I think the schema_id property isn't in line yet. The spec uses snake_case for it, while the anoncreds implementation currently transforms it to camelCase

Let's wait with this change for now. there's still discussion ongoing on what to do. It may stay at camelCase

PR #25, opened 2022-12-07
No id creation

### LINKED ISSUE #4: Update schema methods to not generate id based on schema values
As the identifiers can now be any URI, we should update the methods in the AnonCreds library to not generate the id values, but rather allow the user to generate the IDs themselves based on the AnonCreds method they're using.

- [ ] Update the `credx_create_schema` ffi method based on the chosen approach
    -  remove `origin_did` as parameter
    - remove the `seq_no` parameter from the schema creation (but keep it in the data model for now to not break cred def flow)
    - based on the chosen approach, add an schema_id parameter or not
- [ ] Update the `create_schema` issuer.rs method based on the choses approach
    - remove `origin_did` parameter
    - remove the `seq_no` parameter from the schema creation (but keep it in the data model for now to not break cred def flow)
    - based on the chosen approach, add a schema_id parameter or not
    - remove the generation of the schema_id based on the schema values
- [ ] If the frist approach is taken, we need to update all validation logic and make the schema_id optional (but only in some cases). This requires some refactoring probably

One things to figure out is how we want to approach the generation of the ID, and whether it already needs to be present when we call the methods that create the objects.  There's two approaches we can take:

## Approach 1
Call the creation method (e.g. `create_schema`) _without_ any identifier and then return the created object (the schema) without the `id` property. The schema is now created and the `id` property can be added later when the object is written to the ledger.

The advantage of this appraoch is that it allows the id generation process to be based on the contents of the object (schema), or it allows the id to be known after the object has been written to the ledger (if e.g. the ledger generates some identifier). 

## Approach 2

Call the creation method (e.g. `create_schema`) _with_ the identifier and return the created object (the schema) with the `id` property.

The advantage of this appraoch is that is allows the anoncreds library to validate the identifier to be a valid URI / legacy indy identifier and we don't have a in-between representation of the model (all fields except the id).

### LINKED ISSUE #13: Update credential definition methods to not generate id based on credential definition values
See https://github.com/hyperledger/anoncreds-rs/issues/4 for general context and possible approaches

As the identifiers can now be any URI, we should update the methods in the AnonCreds library to not generate the id values, but rather allow the user to generate the IDs themselves based on the AnonCreds method they're using.

- [ ] Update the `credx_create_credential_definition` ffi method based on the chosen approach
    -  remove `origin_did` as parameter
    - based on the chosen approach, add a cred_def_id parameter or not
- [ ] Update the `create_credential_definition` issuer.rs method based on the choses approach
    - remove `origin_did` parameter
    - based on the chosen approach, add a cred_def_id parameter or not
    - remove the generation of the cred_def_id based on the credential definition values
    - remove the logic of extracting the seq_no from the schema for the schema_id property in the cred_def, but rather use the schema.id directly for the schema_id property.
- [ ] If the frist approach is taken, we need to update all validation logic and make the cred_def_id optional (but only in some cases). This requires some refactoring probably

### LINKED ISSUE #23: Update the revocation registry definition methods to not generate id based on revocation registry values
See https://github.com/hyperledger/anoncreds-rs/issues/4 for general context and possible approaches

As the identifiers can now be any URI, we should update the methods in the AnonCreds library to not generate the id values, but rather allow the user to generate the IDs themselves based on the AnonCreds method they're using.

- [ ] Update the `credx_create_revocation_registry` ffi method based on the chosen approach
    -  remove `origin_did` as parameter
- [ ] Update the `create_revocation_registry` issuer.rs method based on the chosen approach
    - remove `origin_did` parameter
    - remove the generation of the rev_reg_id based on the input values. The model will now be created without an `id` property.
- Remove the `id` property from the revocation registry definition model

blu3beri:
closes #4
closes #13 
closes #23 

It also picks up some tasks from other tickets. Mainly with regards to the "remove to_unqualified...".

Might have to do some clean up (most `String` can be converted to `&str`).

There are two major TODOs which I need some help with. 

Right now we have some presentation filter, like: schema issuer, schema name, etc. Before we could derive this from the schema id, but not anymore. How will we deal with this? As the schema id has no guarantee of being "qualified" we can not use it. 

I believe we can do these from the filters without changing anything:

```rust
    Ok(Filter {
        schema_id,         // YES
        schema_name,       // NO
        schema_issuer_did, // NO
        schema_version,    // NO
        cred_def_id,       // YES
        issuer_did,        // NO
    })
```
We can allow for more data to be passed in which can be gotten from an object resolution on the ledger or whatever.

The second one is the wrappers. I did not update them, but updating them would be trivial as they are just FFI-wrappers without complex logic build atop. I can create an issue for that if there is none.


anoncreds/src/ffi/credential.rs
```diff
@@ -59,6 +59,7 @@ pub extern "C" fn anoncreds_create_credential(
     attr_names: FfiStrList,
     attr_raw_values: FfiStrList,
     attr_enc_values: FfiStrList,
+    rev_reg_id: FfiStr,
```
TimoGlastra, line 62: Why wasn't this needed before?

anoncreds/src/ffi/presentation.rs
```diff
@@ -65,17 +65,34 @@ pub extern "C" fn anoncreds_create_presentation(
     self_attest_values: FfiStrList,
     master_secret: ObjectHandle,
     schemas: FfiList<ObjectHandle>,
+    // TODO: this is index-matched with schemas. Is there a better solution?
+    // since we have an objecthandle list for the schemas we can not add Ids to them
+    // Otherwise they all would need ids and we would like to seperate that
+    // Also, we kind of do the same with `self_attested_names` and `self_attested_values`
+    schema_ids: FfiStrList,
```
TimoGlastra, line 72: In AFJ (and I think the indy-sdk public api) we do it like this:

```
{
  "<schemaId>": <schemaObject>
}
```

Would that work? That way you can't really have them out of sync

anoncreds/src/services/verifier.rs
```diff
@@ -30,10 +30,10 @@ static INTERNAL_TAG_MATCHER: Lazy<Regex> =
 pub fn verify_presentation(
     presentation: &Presentation,
     pres_req: &PresentationRequest,
-    schemas: &HashMap<SchemaId, &Schema>,
-    cred_defs: &HashMap<CredentialDefinitionId, &CredentialDefinition>,
-    rev_reg_defs: Option<&HashMap<RevocationRegistryId, &RevocationRegistryDefinition>>,
-    rev_regs: Option<&HashMap<RevocationRegistryId, HashMap<u64, &RevocationRegistry>>>,
+    schemas: &HashMap<String, &Schema>,
```
TimoGlastra, line 33: Like here @blu3beri, where we have a map of schem_id => schema

anoncreds/tests/anoncreds_demos.rs
```diff
@@ -35,7 +34,7 @@ fn anoncreds_works_for_single_issuer_single_prover() {
 
     // Issuer creates Credential Definition
     let cred_def_parts = issuer::create_credential_definition(
-        &issuer_wallet.did,
+        "SAMPLE_ID".to_owned(),
```
TimoGlastra, line 37: This should fail, as it is not a valid URI -- that's why I would like to keep id validation in place where we use it

anoncreds/src/services/verifier.rs
```diff
@@ -675,28 +675,26 @@ fn gather_filter_info(referent: &str, identifiers: &HashMap<String, Identifier>)
...
-            err_msg!(
-                "Invalid Schema ID `{}`: wrong number of parts",
-                identifier.schema_id.0
-            )
-        })?;
+    // TODO: how can we get these as as we can not extract them from the ID anymore
+    let schema_name = String::from("");
+    let schema_version = String::from("");
+    let schema_issuer_did = String::from("");
+    let cred_def_issuer_did = Some(String::from(""));
```
TimoGlastra, line 682: We can extract the schema_name and schema_version from the passed schemas in the `verify_presentation` method and pass them down (`verify_presentation`  -> `verify_requested_restrictions` -> `gather_filter_info`).

For the other ones, maybe we need to look at adding an `issuerId` property to anoncreds objects after all? I think there's no harm in it, and it would be backwards compatible with current anoncreds (issuerId is always the indy did in that case). Relevant issue: https://github.com/hyperledger/anoncreds-spec/issues/102

For the indy method it would be the indy did, for cheqd it would be the cheqd did. For non ledger based methods there must be some point of reference of who created the schema/credential definition. For https it could be the host (e.g. schema `https://hyperledger.org/membershipSchema` would have `issuerId` of `https://hyperledger.org`

@swcurran thoughts?

TimoGlastra (changes requested):
Woot, great work @blu3beri!!

TimoGlastra:
@dkulic this is probably also relevant to you, as it touches some things you were also working I think

anoncreds/src/ffi/cred_def.rs
```diff
@@ -56,6 +58,29 @@ pub extern "C" fn anoncreds_create_credential_definition(
...
+                CredentialDefinition::CredentialDefinitionV1(cred_def) => {
+                    cred_def.schema_id.to_string()
+                }
+            },
+            s => return Err(err_msg!("Unsupported attribute: {}", s)),
+        };
+        unsafe { *result_p = rust_string_to_c(val) };
+        Ok(())
+    })
+}
```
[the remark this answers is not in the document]
blu3beri: This method was removed by me, but added it back. It can get the schema property from a handle, but I think we can remove it as it does not add any value.

anoncreds/src/ffi/cred_def.rs
```diff
@@ -1,22 +1,20 @@
...
+        CredentialDefinition, CredentialDefinitionConfig, CredentialDefinitionPrivate,
+        CredentialKeyCorrectnessProof as KeyCorrectnessProof, SignatureType,
     },
 };
 
 #[no_mangle]
 pub extern "C" fn anoncreds_create_credential_definition(
-    origin_did: FfiStr,
+    schema_id: FfiStr,
     schema: ObjectHandle,
```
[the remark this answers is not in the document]
blu3beri: Within the library we have an ObjectHandle which points to a mapping between handles and anoncredsobjects.

If we add a schema id to that it would be a weird mapping. Like, this would be a allowed, but quite vague.

```rust
struct AnonCredsobjectWithId {
  id: String,
  object: AnonCredsObject
}
```
But, my preference goes to the seperate schema_id, as it makes the ffi wrapper a lot easier to work with. Nested objects, structs, classes, etc. can get very annoying when sending them over an FFI layer and keeping it simple like this we avoid a lot of IMO unnecessary work.

anoncreds/src/ffi/credential.rs
blu3beri: I believe it was inside the `rev_reg` object

anoncreds/src/ffi/presentation.rs
blu3beri: No, I am not too sure if we can work with (hash)maps over FFI easily. As with the last point, keeping the FFI layer simple types has a very high priority for me. Since they are ordered lists it will be fine, but I am not a fan of it as well :).

anoncreds/src/services/verifier.rs
blu3beri: Yeah internally this works very well and is the best solution.

TimoGlastra (approved):
LGTM, but not a rust developer so maybe someone should take a look at it aswell :)

swcurran:
@andrewwhitehead -- can you take a look, per @TimoGlastra request?

blu3beri:
Just as a note, I will create a follow up PR doing some cleanup for the codebase.

TimoGlastra:
@whalelephant feel free to also leave a review if you have time

dkulic:
Wow, great work. Thanks.
I was busy with some other stuff, will try to get involved more.

anoncreds/src/services/verifier.rs
andrewwhitehead: I agree it makes sense to attach an issuer ID to the schema and cred def objects. While the ID of the object might not be known prior to publishing, the issuer ID should always be known.

anoncreds/src/data_types/anoncreds/cred_def.rs
```diff
@@ -76,13 +57,6 @@ impl CredentialDefinitionV1 {
     }
 }
 
-impl Validatable for CredentialDefinitionV1 {
-    fn validate(&self) -> Result<(), ValidationError> {
-        self.id.validate()?;
-        self.schema_id.validate()
```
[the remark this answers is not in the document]
andrewwhitehead: I'm in favour of having a String wrapper type for identifiers (ObjectID?) with validation rules to accept only valid URIs (maybe not every aspect of validating URIs, but the basics).

anoncreds/src/ffi/cred_def.rs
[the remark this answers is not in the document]
andrewwhitehead: It's used by the python wrapper, and avoids the caller having to unwrap different versions of a credential definition to find the property (although at the moment there's only one version). There's a similar method for revocation registry definitions that supports more fields: https://github.com/hyperledger/anoncreds-rs/blob/38e9a55c379694112b04c9def4ad3807e79bbe4c/anoncreds/src/ffi/revocation.rs#L172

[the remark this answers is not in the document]
andrewwhitehead: Incidentally I'm not sure there's any benefit to supporting multiple 'versions', it just complicates the code everywhere, although I wouldn't rip it out without some consideration.

anoncreds/src/data_types/anoncreds/cred_def.rs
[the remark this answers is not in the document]
andrewwhitehead: The wrapper type could even contain something like `Arc<str>` to make cloning cheaper.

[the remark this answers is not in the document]
blu3beri: I don't think that we handle it between threads, so we an `Rc<T>` would suffice. You will likely know it better, though so please correct me if I am wrong.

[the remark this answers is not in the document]
andrewwhitehead: It could be a bit limiting to use a type that isn't `Send`. Also a String is probably just fine on consideration.

anoncreds/src/services/verifier.rs
blu3beri: Would it be good to pick this up when the specification is updated? Seems like an annoying blocker and for now we could merge it with the TODO. I can create an issue for it to resolve it.

anoncreds/src/data_types/anoncreds/credential.rs
```diff
@@ -70,8 +67,8 @@ pub struct CredentialInfo {
     pub referent: String,
     pub attrs: ShortCredentialValues,
     pub schema_id: SchemaId,
-    pub cred_def_id: CredentialDefinitionId,
-    pub rev_reg_id: Option<RevocationRegistryId>,
+    pub cred_def_id: String,
```
dkulic, line 70: Is there any reason why we do not use CredentialDefinitionId here?

anoncreds/src/data_types/anoncreds/macros.rs
```diff
@@ -0,0 +1,62 @@
...
+            pub fn validated_new(
+                s: impl Into<String>,
+            ) -> Result<Self, crate::data_types::ValidationError> {
+                let s = Self(s.into());
+                s.validate()?;
+                Ok(s)
+            }
+        }
+
+        impl crate::data_types::Validatable for $i {
```
dkulic, line 21: It confuses me that we have this validation for id data types implemented, but all uses are removed, as far as I can tell?

blu3beri:
I used quite a bit of `impl Into<SchemaId>` but it might be better to do `impl TryInto<SchemaId>` so we can validate when we call into. This almost automatically adds validation. Can do this in a follow up PR to keep this one smaller.

dkulic (approved):
LGTM

anoncreds/src/ffi/cred_def.rs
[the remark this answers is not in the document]
blu3beri: I removed the V1 versions in another PR.

anoncreds/src/services/verifier.rs
blu3beri: I created an issue for this, will fix this once we have a concrete answer in the spec.

PR #27, opened 2022-12-10
Update to idiomatic rust

blu3beri:
Dependent on #25 
- Resolved some issues that clippy was giving. 
- renamed some functions to not make them collide with functions from std

Also fixed some small stuff from indy-utils, just to make clippy be quiet.


anoncreds/src/data_types/anoncreds/cred_def.rs
```diff
@@ -1,17 +1,19 @@
...
+
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub enum SignatureType {
     CL,
 }
 
 impl SignatureType {
-    pub fn from_str(value: &str) -> Result<Self, ConversionError> {
-        match value {
+    pub fn from_string(value: impl Into<String>) -> Result<Self, ConversionError> {
```
andrewwhitehead, line 15: It might be more idiomatic to implement the FromStr trait instead of an inherent method. This change allocates a new String unnecessarily if an `&str` is passed in.

anoncreds/src/data_types/anoncreds/master_secret.rs
```diff
@@ -14,7 +14,7 @@ impl MasterSecret {
     #[inline]
     pub fn new() -> Result<Self, ConversionError> {
         let value = UrsaProver::new_master_secret().map_err(|err| {
-            ConversionError::from_msg(format!("Error creating master secret: {}", err))
+            ConversionError::from_msg(format!("Error creating master secret: {err}"))
```
andrewwhitehead, line 17: NB this requires Rust 1.58 minimum. I think that's reasonable but should be added to Cargo.toml. 2021 edition requires 1.56 minimum so we can update to that as well.

anoncreds/src/services/issuer.rs
```diff
@@ -291,24 +215,20 @@ pub fn update_revocation_registry(
 }
 
 pub fn create_credential_offer(
-    schema_id: &SchemaId,
-    cred_def: &CredentialDefinition,
+    schema_id: impl Into<SchemaId>,
+    cred_def_id: &str,
```
andrewwhitehead, line 219: Not sure why one of these is an `Into<SchemaId>` and the other a `&str`?

andrewwhitehead: Also I don't love `Into<SchemaId>` because it skips validation (can't return an Err).

blu3beri: My idea was to internally use the SchemaId and for the API we use `impl Into<T>`. when we get the schema_id we call it like so:

```rust
fn test(s: impl Into<SchemaId>) -> Result<Any> {
  let s = schema_id.into().validate()?;
}
```
This way we always validate when the value comes in, but we do not require the user to cast every string into a `SchemaId`.

blu3beri: It is incorrect that it is inconsistent, this should be fixed.

anoncreds/src/data_types/anoncreds/master_secret.rs
blu3beri: Yeah I can add 2021 to this PR.

anoncreds/src/data_types/anoncreds/cred_def.rs
blu3beri: yeah the reason that I did not do that is that from_str does not return a Result. TryFrom would be good here as I assume it does return a result.

andrewwhitehead: FromStr::from_str does result a Result, like TryFrom

blu3beri: Oh yeah, huh. I tried to fix this issue with implementing it on FromStr, but I think I did From<str>. Do you have a preference for TryFrom or FromStr?

andrewwhitehead: I guess my preference is FromStr, although it does require an include when used.

anoncreds/src/services/issuer.rs
andrewwhitehead: It does seem more idiomatic to me for it to be a validating wrapper, ie. validation is performed when it is constructed rather than potentially being missed later. The struct(s) could have a `const fn new_unchecked(value: &str)` for skipping validation in tests or const contexts.

blu3beri: Agreed. I created a `validated_new()` (did not see this comment before that) which does what your implementation of `new` would do. I think that validating with `new` is better, as it will validate by default.

I also left a comment at #25 where I propose TryInto instead of Into, which makes a lot of sense here IMO. If that seems okay with you, I can create a follow up PR where Into will be changed to TryInto.

blu3beri: Implemented `new` with validation in #25 .

anoncreds/src/ffi/mod.rs
```diff
@@ -10,7 +10,7 @@ ffi_support::define_string_destructor!(anoncreds_string_free);
 #[no_mangle]
 pub extern "C" fn anoncreds_buffer_free(buffer: ByteBuffer) {
     ffi_support::abort_on_panic::with_abort_on_panic(|| {
-        drop(buffer.destroy_into_vec().zeroize());
+        buffer.destroy_into_vec().zeroize();
```
blu3beri, line 13: I assume we do not have to call drop here ourselves, right?

blu3beri:
I converted all the `Into<T>` to `TryInto<T>` but a downside now is that if we request `Option<impl Into<SchemaId>`, and we want to supply `None`, we MUST do it like this: `None::<SchemaId>` which I do not like. We only do this for the rev_reg_def_id, but it is a weird inconsistency. Not too sure what the right direction is here.

TimoGlastra (approved):

PR #31, opened 2022-12-12
Update schema data model

### LINKED ISSUE #10: Update the Schema data type according to the AnonCreds specification
The data model has been slightly adjusted to be more generic.

The data model is described here: [anoncreds-wg.github.io/anoncreds-spec/#schema-publisher-publish-schema-object](https://anoncreds-wg.github.io/anoncreds-spec/#schema-publisher-publish-schema-object). We diverge a bit to keep it consistent with the current implementation

```
{
    "id": "https://www.did.example/schema.json",
    "name": "Example schema",
    "version": "0.0.1",
    "attrNames": ["name", "age", "vmax"]
}
```

- id - (string) The identifier of the [Schema](https://anoncreds-wg.github.io/anoncreds-spec/#term:schema). The format of the identifier is dependent on the [AnonCreds Objects Method]([anoncreds-wg.github.io/anoncreds](https://anoncreds-wg.github.io/anoncreds) spec/#term:anoncreds-objects-method) used in publishing the [Schema](https://anoncreds-wg.github.io/anoncreds-spec/#term:schema).
- name (string) - the name of the schema
- version (string) - the schema version
- attrNames (str[]) - an array of strings with each string being the name of an attribute of the schema

This results in the following tasks:
-  Keep the seq_no property in for now to not break the cred def flow. This will be removed in a separate task
 Update the serde encoding of attrNames property to attr_names (maybe we should keep it attrNames to not change unneeded values?)
-  Update Validatable for SchemaV1 to only validate the schema name and version against the identifier if is_legacy_indy_identifier is true on self.id
- Remove the to_unqualified method, there's no need to be switching between qualified/unqualified identifiers types, we just use the identifiers as is

blu3beri:
Dependent on #23, #27 and #30

- Removes the ver from the schema
- Rename attrNames to attr_names
- remove internal SchemaV1 enum wrapper

closes #10 

### DIFF
--- a/anoncreds/src/data_types/anoncreds/schema.rs
+++ b/anoncreds/src/data_types/anoncreds/schema.rs
@@ -8,21 +8,13 @@ pub const MAX_ATTRIBUTES_COUNT: usize = 125;
 
 impl_anoncreds_object_identifier!(SchemaId);
 
-#[derive(Clone, Debug, Serialize, Deserialize)]
-#[serde(tag = "ver")]
-pub enum Schema {
-    #[serde(rename = "1.0")]
-    SchemaV1(SchemaV1),
-}
-
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct SchemaV1 {
+pub struct Schema {
     pub name: String,
     pub version: String,
     #[serde(rename = "attrNames")]
     pub attr_names: AttributeNames,
-    pub seq_no: Option<u32>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
@@ -56,7 +48,7 @@ impl From<AttributeNames> for HashSet<String> {
     }
 }
 
-impl Validatable for SchemaV1 {
+impl Validatable for Schema {
     fn validate(&self) -> Result<(), ValidationError> {
         self.attr_names.validate()?;
         Ok(())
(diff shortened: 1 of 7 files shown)


anoncreds/src/data_types/anoncreds/schema.rs
```diff
@@ -1,67 +1,23 @@
...
+impl_anoncreds_object_identifier!(SchemaId);
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase")]
-pub struct SchemaV1 {
-    pub id: SchemaId,
+pub struct Schema {
     pub name: String,
     pub version: String,
-    #[serde(rename = "attrNames")]
```
TimoGlastra, line 51: Let's wait with this one for a bit maybe. I know I created an issue to change it. But as all models are coming out to use camelCase, I think we should just jeep using camelCase here and minimize changes.

Thoughts?

blu3beri: I am fine with either. I think a definitive decision would be very valuable here. 

@swcurran I might have missed some of the on going discussions surrounding the spec, but is there a decision made around the anoncreds objects and which casing they use? I think I have a slight preference for snake case, but without any good reason, so either is good in my opinion. (if this needs further discussion, and there is no issue open already, I would not mind creating one).

TimoGlastra: The reason I'm leaning to keeping it camelCase is that anoncreds objects already mostly use camelCase. The records that are registered on the indy ledger are using snake_case in some places however.

So my opinion would be to go with camelCase

TimoGlastra (commented):
LGTM, expect that I would like to keep the attr_names as camelCase for now. I just opened a Pr in the spec today to update it to camelCase

TimoGlastra:
> LGTM, expect that I would like to keep the attr_names as camelCase for now. I just opened a Pr in the spec today to update it to camelCase

If the discussion ends up we will use snake_case for everything then we can address in a separate PR

TimoGlastra (approved):

PR #35, opened 2022-12-15
add comment about openssl dependency

blu3beri:
@andrewwhitehead I am not sure why it was a dependency, but if we have to keep it for some reason we can of course keep it in.

Signed-off-by: blu3beri <berend@animo.id>


blu3beri:
Ah I see now why we use it. It is because ursa does not have a vendored openssl feature open. I will add a comment explaining and reuse this PR for that.

TimoGlastra (approved):

PR #40, opened 2023-01-03
Integrate indy-utils into anoncreds

### LINKED ISSUE #33: Remove / integrate indy-utils
Right now we still have the `indy-utils` package which is partially being used. 
- [ ] Do we want to create a anoncreds-utils package or bring them inside the anoncreds folder? 
- [ ] Which modules can we keep inside the anoncreds package?
- [ ] Which of the crypto packages to we want to bring?
- [ ] How can we keep the implementation crypto agile (where it makes sense)?
- [ ] Do we need to keep the did module?

blu3beri:
closes #33 

- Removes unused packages from the `indy-utils` package and merged the used packages within anoncreds to the `utils` directory.
- Removed the cargo workspace as we only have one project.
- removed `data_types/anoncreds/...` directory as the only data types we will have are `anoncreds`.


src/utils/validation.rs
```diff
@@ -0,0 +1,38 @@
...
+    ($($arg:tt)+) => {
+        $crate::error::ValidationError::from(format!($($arg)+))
+    };
+}
+
+/// Trait for data types which need validation after being loaded from external sources
+/// TODO: this should not default to Ok(())
+pub trait Validatable {
+    fn validate(&self) -> Result<(), ValidationError> {
+        Ok(())
```
TimoGlastra, line 36: Should we open an issue to not forget?

TimoGlastra (approved):
LGTM. Maybe @whalelephant can also take a look?

src/utils/validation.rs
blu3beri: Yeah probably. This was done in indy-credx for items that might get future validation (not 100% sure though). So we might be able to just remove the validation on objects that do not need validation.

PR #44, opened 2023-01-04
build: Add build for ios and android

### LINKED ISSUE #45: Create xcodeframework for anoncreds React Native

### LINKED ISSUE #46: Create android-libs folder for anoncreds React native

blu3beri:
Depends on #43 

closes #45 
closes #46 

- I had to patch openssl-src-rs, nothing functional, https://github.com/alexcrichton/openssl-src-rs/compare/release/111...blu3beri:openssl-src-rs:release/111 as it did not provide a build script for the target `aarch64-apple-ios-sim`. Since they do not accept non-security fixes for this branch we can not get this upstream, unless we depend upon openssl 3.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,5 +9,10 @@ members = [
 lto = true
 codegen-units = 1
 
+# openssl-src does not have a build script for the target `aarch64-apple-ios-sim`, which we need.
+# This fork has that target added and no functional changes. There is a pull request open for
+# the fix but previous pull requests were closed with the same context as they were not security
+# fixes. When this pull request is merged, we can depend on the `upstream` git repo or, when
+# released, we can depend on the latest version of `openssl-src`.
 [patch.crates-io]
 openssl-src = { git = "https://github.com/blu3beri/openssl-src-rs", branch = "release/111" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,3 +8,6 @@ members = [
 [profile.release]
 lto = true
 codegen-units = 1
+
+[patch.crates-io]
+openssl-src = { git = "https://github.com/blu3beri/openssl-src-rs", branch = "release/111" }
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,14 +1,14 @@
-name: "Build Packages"
+name: 'Build Packages'
 
-"on":
+'on':
   release:
     types: [created]
   workflow_dispatch:
     inputs:
       publish:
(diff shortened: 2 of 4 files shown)


TimoGlastra (approved):
Where does it mention they only accept security PRs? Not sure what to do if we can't get it in. This hasn't been solved for indy-credx yet a while ago?

Cargo.toml
```diff
@@ -8,3 +8,6 @@ members = [
 [profile.release]
 lto = true
 codegen-units = 1
+
+[patch.crates-io]
+openssl-src = { git = "https://github.com/blu3beri/openssl-src-rs", branch = "release/111" }
```
TimoGlastra, line 13: Maybe add a comment here why it's patched?

blu3beri:
> > Where does it mention they only accept security PRs? Not sure what to do if we can't get it in. This hasn't been solved for indy-credx yet a while ago?
> 
> Here for example: https://github.com/alexcrichton/openssl-src-rs/pull/130#issuecomment-1069625131
> 
> I don't think it was fixed for credx, I vaguely remember the issue from before. Would be nice if we can just have a patch file in this repo with the fix. Can check for that later on. 

However, this works for now.

Cargo.toml
TimoGlastra: > > Where does it mention they only accept security PRs? Not sure what to do if we can't get it in. This hasn't been solved for indy-credx yet a while ago?
> 
> Here for example: [alexcrichton/openssl-src-rs#130 (comment)](https://github.com/alexcrichton/openssl-src-rs/pull/130#issuecomment-1069625131)
> 
> I don't think it was fixed for credx, I vaguely remember the issue from before. Would be nice if we can just have a patch file in this repo with the fix. Can check for that later on.

We may be able to say it IS pressing as otherwise we'll have forks that don't necesarily improve security 🤷

TimoGlastra:
What would be the effort to upgrade to openssl 3?

blu3beri:
> What would be the effort to upgrade to openssl 3?

It is coming from `Ursa`, so if it is not just updating the dependency, I don't think I would be able to fix it quickly.

Cargo.toml
blu3beri: Yeah I will create a PR with it. If it is merged I will remove the fix but leave this in for now.

TimoGlastra:
> @TimoGlastra Can we merge this? I think the comments are resolved.

Go for it

PR #49, opened 2023-01-06
Return NRP warning if presentation timestamp is outside of req interval

### LINKED ISSUE #41: Unchecked revocation interval between Presentation and PresRequest
The `NonRevocationInterval` in the `PresentationRequest` on  the request or the attribute/predicate level are not compared with the timestamp of which the prover has updated the `RevocationState` to in creating the `Presentation`. 

Given the verifier only provides `RevocationRegistry` (aka Accumulator value)  for the timestamps in the `NonRevocationInterval`, the required `RevocationRegistry` for the timestamp given in the `Presentation` will be missing and would not cause a revoked credential during that interval to be verified. 
However, if the verifier provides `RevocationRegistry` mapped to timestamps outside of the interval,
a revoked credential can be verified. 

This might also be related to the issued in #36, i.e. if the `RevocationState` is for a timestamp that is outside of the one defined in the `PresentationRequest`, then regardless if it has been revoked, the verifier might not supply the Accumulator value for such a timestamp.

whalelephant:
Closes #41

### DIFF
--- a/anoncreds/src/data_types/anoncreds/rev_reg.rs
+++ b/anoncreds/src/data_types/anoncreds/rev_reg.rs
@@ -150,6 +150,28 @@ pub mod serde_revocation_list {
     }
 }
 
+// NonRevocationProof warnings:
+// - timestamps_out_of_range: all timestamps used for rev_state that are o.o.r
+//   this can be for request / attribute / predicate level time intervals
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub struct NonRevProofWarnings {
+    pub timestamps_out_of_range: Option<Vec<String>>,
+}
+
+impl NonRevProofWarnings {
+    pub fn new(referents: Vec<String>) -> Self {
+        if referents.is_empty() {
+            Self {
+                timestamps_out_of_range: None,
+            }
+        } else {
+            Self {
+                timestamps_out_of_range: Some(referents),
+            }
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/anoncreds/src/services/verifier.rs
+++ b/anoncreds/src/services/verifier.rs
@@ -7,7 +7,7 @@ use super::helpers::*;
 use super::types::*;
 use crate::data_types::anoncreds::cred_def::CredentialDefinition;
 use crate::data_types::anoncreds::cred_def::CredentialDefinitionId;
-use crate::data_types::anoncreds::rev_reg::RevocationRegistryId;
+use crate::data_types::anoncreds::rev_reg::{NonRevProofWarnings, RevocationRegistryId};
 use crate::data_types::anoncreds::rev_reg_def::RevocationRegistryDefinitionId;
 use crate::data_types::anoncreds::schema::Schema;
 use crate::data_types::anoncreds::schema::SchemaId;
@@ -40,7 +40,7 @@ pub fn verify_presentation(
     cred_defs: &HashMap<&CredentialDefinitionId, &CredentialDefinition>,
     rev_reg_defs: Option<&HashMap<&RevocationRegistryDefinitionId, &RevocationRegistryDefinition>>,
     rev_regs: Option<&HashMap<RevocationRegistryId, HashMap<u64, &RevocationRegistry>>>,
-) -> Result<bool> {
+) -> Result<(bool, NonRevProofWarnings)> {
     trace!("verify >>> presentation: {:?}, pres_req: {:?}, schemas: {:?}, cred_defs: {:?}, rev_reg_defs: {:?} rev_regs: {:?}",
     presentation, pres_req, schemas, cred_defs, rev_reg_defs, rev_regs);
 
@@ -51,6 +51,7 @@ pub fn verify_presentation(
         received_unrevealed_attrs(presentation)?;
     let received_predicates: HashMap<String, Identifier> = received_predicates(presentation)?;
     let received_self_attested_attrs: HashSet<String> = received_self_attested_attrs(presentation);
+    let mut nrp_warnings = NonRevProofWarnings::default();
 
     compare_attr_from_proof_and_request(
         pres_req,
@@ -79,6 +80,7 @@ pub fn verify_presentation(
         &received_unrevealed_attrs,
         &received_self_attested_attrs,
         &received_predicates,
+        &mut nrp_warnings,
     )?;
 
     let mut proof_verifier = CryptoVerifier::new_proof_verifier()?;
@@ -191,8 +193,9 @@ pub fn verify_presentation(
     let valid = proof_verifier.verify(&presentation.proof, pres_req.nonce.as_native())?;
 
     trace!("verify <<< valid: {:?}", valid);
+    trace!("non_rev_proof_warnings <<<: {:?}", nrp_warnings);
 
-    Ok(valid)
+    Ok((valid, nrp_warnings))
 }
 
 pub fn generate_nonce() -> Result<Nonce> {
@@ -314,12 +317,15 @@ fn compare_timestamps_from_proof_and_request(
     received_unrevealed_attrs: &HashMap<String, Identifier>,
     received_self_attested_attrs: &HashSet<String>,
     received_predicates: &HashMap<String, Identifier>,
+    nrp_warnings: &mut NonRevProofWarnings,
 ) -> Result<()> {
+    let mut warnings = vec![];
     pres_req
         .requested_attributes
         .iter()
         .map(|(referent, info)| {
             validate_timestamp(
+                &mut warnings,
                 received_revealed_attrs,
                 referent,
                 &pres_req.non_revoked,
@@ -327,6 +333,7 @@ fn compare_timestamps_from_proof_and_request(
             )
             .or_else(|_| {
                 validate_timestamp(
+                    &mut warnings,
                     received_unrevealed_attrs,
                     referent,
                     &pres_req.non_revoked,
@@ -337,7 +344,7 @@ fn compare_timestamps_from_proof_and_request(
                 received_self_attested_attrs
                     .get(referent)
                     .map(|_| ())
-                    .ok_or_else(|| err_msg!("Missing referent"))
+                    .ok_or_else(|| err_msg!("Referent validation failed: {}", referent))
             })
         })
         .collect::<Result<Vec<()>>>()?;
@@ -347,6 +354,7 @@ fn compare_timestamps_from_proof_and_request(
         .iter()
         .map(|(referent, info)| {
             validate_timestamp(
+                &mut warnings,
                 received_predicates,
                 referent,
                 &pres_req.non_revoked,
@@ -354,6 +362,10 @@ fn compare_timestamps_from_proof_and_request(
             )
         })
         .collect::<Result<Vec<()>>>()?;
+
+    if !warnings.is_empty() {
+        nrp_warnings.timestamps_out_of_range = Some(warnings);
+    }
 
     Ok(())
 }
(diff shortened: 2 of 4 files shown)


TimoGlastra (commented):
It's great that we're expanding the return type of the verify proof method. However, I'm not a fan of keep the default isVerified `true` and seeing the NRP as a warning. If there's a proof request with a certain requirement, and that requirement is not met, is the proof still valid? Some parts may be, but I'm not sure if we should return a single boolean in that case. It's making it too easy to mess things up, and can give the wrong impression.

I quite like how Trinsic has solved this issue in their API, and maybe we can make it similair to this. The structure of their verify proof call looks like this:

```json
{
  "is_valid": true,
  "validation_results": {
    "SignatureVerification": {
        "is_valid": true,
         "messages": [],
    },
    "CredentialStatus": {
        "is_valid": true,
         "messages": [],
    },
    "SchemaConformance": {
        "is_valid": true,
         "messages": [],
    },
    "TrustRegistryMembership": {
        "is_valid": true,
         "messages": [],
    },
    "IssuerIsSigner": {
        "is_valid": true,
         "messages": [],
    },
  }
}
```

Each of the `validation_results` referes to a separate validation step that is taken. Only if all `validation_results.is_valid` evaluate to `true`, the top level `is_valid` is true. 

In this case the `SignatureVerification` will be valid, but the `CredentialStatus` (or whatever we want to call it in AnonCreds) is not valid according to the proof request.

cc @swcurran

swcurran:
In general I like that. However, the problem I'm highlighting with the interval is that we don't (and, I would argue) can't know if the interval requirement has been satisfied for the reasons given on the other issue.  If both parties had a list of all of the RevocationStatusList entry timestamps, it would be possible, but I think there will be too many occasions when a verification will be incorrectly flagged as false because, the timestamp used is outside of the interval.   I think we should look for a better solution in V2.0, but in v1.0, I think it is tricky.

To be precise -- if a verifier gives the interval '{ "from": "now()", "to": "now()" }`, as we currently recommend that verifiers do, the presentation will always be invalid (unless of course, the issuer happens to publish a new revocation registry at exactly "now()"). The holder creating an NRP based on the RevocationStatusList entry immediately before the interval is technically "within the interval", but that is very difficult for the AnonCreds verifier code to check. 

The semantics of interval are too complex to properly implement if there is no way to query the ledger to know what timestamped transactions are within the interval.

anoncreds/src/data_types/anoncreds/rev_reg.rs
```diff
@@ -150,6 +150,28 @@ pub mod serde_revocation_list {
...
+pub struct NonRevProofWarnings {
+    pub timestamps_out_of_range: Option<Vec<String>>,
+}
+
+impl NonRevProofWarnings {
+    pub fn new(referents: Vec<String>) -> Self {
+        if referents.is_empty() {
+            Self {
+                timestamps_out_of_range: None,
+            }
```
blu3beri, line 166:
```suggestion
            Self::default()
```

anoncreds/src/services/verifier.rs
```diff
@@ -337,7 +344,7 @@ fn compare_timestamps_from_proof_and_request(
                 received_self_attested_attrs
                     .get(referent)
                     .map(|_| ())
-                    .ok_or_else(|| err_msg!("Missing referent"))
+                    .ok_or_else(|| err_msg!("Referent validation failed: {}", referent))
```
blu3beri, line 347:
```suggestion
                    .ok_or_else(|| err_msg!("Referent validation failed: {referent}"))
```

I believe clippy complains about this now.

anoncreds/src/services/verifier.rs
```diff
@@ -355,6 +363,10 @@ fn compare_timestamps_from_proof_and_request(
         })
         .collect::<Result<Vec<()>>>()?;
 
+    if !warnings.is_empty() {
+        nrp_warnings.timestamps_out_of_range = Some(warnings);
+    }
+
```
blu3beri, line 369: Would it be worth it to instantiate a new one, shadowing the previous one, and keeping `timestamps_out_of_range` private? Not sure if it would work.

anoncreds/src/services/verifier.rs
```diff
@@ -370,26 +382,29 @@ fn compare_timestamps_from_proof_and_request(
 // which was added by the prover when creating `PresentCredentials`,
 // an arg for `create_presentation`.
 //
-// TODO: this timestamp should be compared with the provided interval
+// If the NRP is given at a timestamp outside of the request interval,
+// a wanring is provided for the verifier to decide if they will reject the proof or not
```
blu3beri, line 386:
```suggestion
// a warning is provided for the verifier to decide if they will reject the proof or not
```

anoncreds/src/services/verifier.rs
```diff
@@ -370,26 +382,29 @@ fn compare_timestamps_from_proof_and_request(
...
-        .map(|attr| attr.timestamp.is_some())
-        .unwrap_or(false)
-    {
-        return Err(err_msg!("Missing timestamp"));
+    if let Some(interval) = get_non_revoc_interval(global_interval, local_interval) {
+        // If there is global or local interval, we compare the timestamp provided
+        if let Some(Some(ts)) = received_.get(referent).map(|attr| attr.timestamp) {
+            if ts.gt(&interval.to.unwrap_or(u64::MAX)) || ts.lt(&interval.from.unwrap_or(0)) {
+                // We add to NonRevProofWarnings that the referent is out of range
+                warnings.push(referent.to_string());
```
blu3beri, line 399: Very nit picky but I prefer `to_owned` as it descibes the intention a bit better.

anoncreds/src/services/verifier.rs
```diff
@@ -370,26 +382,29 @@ fn compare_timestamps_from_proof_and_request(
...
+            }
+            Ok(())
+        } else {
+            Err(err_msg!("Missing timestamp"))
+        }
+    } else {
+        Ok(())
```
blu3beri, line 406:
```suggestion
        } else {
            Err(err_msg!("Missing timestamp"))
        }
    Ok(())
```


That would work, right?

anoncreds/src/services/verifier.rs
```diff
@@ -1197,7 +1214,25 @@ mod tests {
             },
         );
         res.insert(
-            "referent_2".to_string(),
+            "referent_before_from".to_string(),
+            Identifier {
+                timestamp: Some(1),
+                schema_id: SchemaId::default(),
```
blu3beri, line 1220: Huh, did not know I implemented default for the anoncreds ids. Ah, could be nice for testing :).

blu3beri (changes requested):
Couple of smaller things, but overall looks good! I am not too familiar with the spec and the correctness of the implementation, so the review is mainly about the code.

TimoGlastra:
> In general I like that. However, the problem I'm highlighting with the interval is that we don't (and, I would argue) can't know if the interval requirement has been satisfied for the reasons given on the other issue. If both parties had a list of all of the RevocationStatusList entry timestamps, it would be possible, but I think there will be too many occasions when a verification will be incorrectly flagged as false because, the timestamp used is outside of the interval. I think we should look for a better solution in V2.0, but in v1.0, I think it is tricky.
> 
> To be precise -- if a verifier gives the interval '{ "from": "now()", "to": "now()" }`, as we currently recommend that verifiers do, the presentation will always be invalid (unless of course, the issuer happens to publish a new revocation registry at exactly "now()"). The holder creating an NRP based on the RevocationStatusList entry immediately before the interval is technically "within the interval", but that is very difficult for the AnonCreds verifier code to check.
> 
> The semantics of interval are too complex to properly implement if there is no way to query the ledger to know what timestamped transactions are within the interval.

Makes sense. I misunderstood the issue. So what we should actually check is whether the timestamp that the proof is based on, is the most accurate based on the proof request? So If I request `now()` and that evaluates to 200, and there's two timestamps available `10` and `190`, we should pick `190`. If there's `10`, `190`, and `210` we should pick either `190` and `210`? 

I still think we should be careful in seeing this as a warning. Ideally we let the verifier provide the closest 'state' (and thus the one that matches with the one the holder provided -- if all is good). This could be tricked however as currently you just provide a map of revocation registries, with the timestamps as keys. Multiple timestamps could be present for the same reovcatino registry for different parts of the proof.  The verifier would need to provide the revocation registry with timestamp per group in the request, so we can throw an error if the revocation registry timestamp the holder used does not match with the revocation registry timestamp the verifier expects the holder to use for a given timeframe

The steps for the verifier when receiving a proof will be:
1. Extract the timestamp the holder used for a specific group name / credential from the proof
2. Fetch the revocation state at the given timestamp from the ledger
3. If the timestamp is within the revocation interval all good and use the revocation registry in the verify method
4. If the timestamp is not within the revocation interval, check if the timestamp the holder used to create the proof is the most logical based on available updates to the ledger. Or, just accept the timestamp the user provided, if you're willing to accept it.

Which of these steps are to be peformed by the upper level vs AnonCreds I'm not sure, but we should maybe add wording to the spec to note these edge cases?

swcurran:
Regards from @TimoGlastra:

> Which of these steps are to be performed by the upper level vs AnonCreds I'm not sure, but we should maybe add wording to the spec to note these edge cases?

I suggest that we do exactly what is in the implementation today (since that is what v1.0 of the specification is intending -- and we're only making changes for ledger-agnosticism and clarification), and we document (as you say) the rough edges in the 1.0 specification.  The problem we have today is that there is not a good way to query the ledger to determine the edges of the interval.

An idea for a what might be a relatively easy clean up is that we update the AnonCreds methods to add to the `resolve_RevReg(revRegId)` response some extra data -- "previous_timestamp", "next_timestamp" (and perhaps "latest_timestamp") so that the verifier can determine if the NRP meets the interval requirement.  Not for today, but perhaps in the next version?

TimoGlastra:
Have thought about it more, and also discussed this with @whalelephant. I do not agree we should return that the proof is verified. If the timestamp is outside of the requested revocation interval extra processing is required by the verifier to determine if it would like to accept the submitted timestamp (based on how close it is to the requested interval, or what timestamps are available on the ledger).

So I'll still like to suggest we return isVerifeid is false, and we add the context object with true/false for each of the specific checks that has been done.  We can then inform the user of this library that if the proof falls outside of the revocation interval, it means all checks will be true, except for one. If that is the case, the verifier can do some additional checks and still see it as valid, but by default I would say it is not valid. If people see isValid=false they'll dig into why it is invalid, if it is true, they'll won't look further

swcurran:
How does that approach compare with what is done in the indy-sdk and credx?

swcurran:
This change will break any verifiers that followed the "Best Practices" from Aries by using the same value for `to` and `from` and did not search for a valid timestamp.  AFAIK, most verifiers use `now()` for those two values, and this change will mean that all of those presentations will fail validation.

TimoGlastra:
Anoncreds rs is a library, and it's still up to the code/framework that is calling this library how to interpret the result.

Maybe it'll force people that update to anoncreds rs to reconsider their current implementstion, or in ACA-Py we could still choose to see it as valid. What is important is that we don't see it as valid by default and let it to the user of this library to decide whether they would like to accept this use case. Maybe it'll break some flows that used to work, but if any timestamp was accepted before it means we now have improved security.

TimoGlastra:
> or in ACA-Py we could still choose to see it as valid.

This is the main point of my comment. AnonCreds is just a library that gives context. ACA-Py decides how to interpret that context.

swcurran:
Let's discuss on Monday in person.  

Still interested to know this:  How does that approach compare with what is done in the indy-sdk and credx?

I'm also interested to know -- if the latest RevStatusList entry is at time T and that is timestamp used by the holder, and the interval time from the verifier was `{From, to = T+t}`, would the approach you are proposing say "verified" or "not verified"?  The correct answer is "verified", but I don't think AnonCreds has enough information to know that.

For the discuss on Monday, I propose that we leave interval as is for v1.0, and we define a new approach for requesting/verifying revocation in the next release.

whalelephant:
Addressed in #102

PR #50, opened 2023-01-09
Added `issuerId` to the `schema` and `cred_def` anoncreds objects

### LINKED ISSUE #34: We can extract the schema_name and schema_version from the passed schemas in the `verify_presentation` method and pass them down (`verify_presentation`  -> `verify_requested_restrictions` -> `gather_filter_info`).
We can extract the schema_name and schema_version from the passed schemas in the `verify_presentation` method and pass them down (`verify_presentation`  -> `verify_requested_restrictions` -> `gather_filter_info`).

For the other ones, maybe we need to look at adding an `issuerId` property to anoncreds objects after all? I think there's no harm in it, and it would be backwards compatible with current anoncreds (issuerId is always the indy did in that case). Relevant issue: https://github.com/hyperledger/anoncreds-spec/issues/102

For the indy method it would be the indy did, for cheqd it would be the cheqd did. For non ledger based methods there must be some point of reference of who created the schema/credential definition. For https it could be the host (e.g. schema `https://hyperledger.org/membershipSchema` would have `issuerId` of `https://hyperledger.org`

- [ ] support both `_did` and `id` for restrictions

@swcurran thoughts?

_Originally posted by @TimoGlastra in https://github.com/hyperledger/anoncreds-rs/pull/25#discussion_r1043001139_

blu3beri:
Partially closes #34 

- Adds `IssuerId` property to the `Schema` and `CredentialDefinition`.

This is reflects the PR for adding the `issuerId` to anoncreds objects.

(revocation needs to be done but is omitted here for conflicts with #37.


anoncreds/src/data_types/anoncreds/cred_def.rs
```diff
@@ -33,12 +33,15 @@ pub struct CredentialDefinitionData {
...
 
 #[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
 pub struct CredentialDefinition {
     pub schema_id: SchemaId,
     #[serde(rename = "type")]
     pub signature_type: SignatureType,
     pub tag: String,
     pub value: CredentialDefinitionData,
+    #[serde(rename = "issuerId")]
```
TimoGlastra, line 43: This is not needed right, as you also have camelCase rename on the struct?

anoncreds/src/services/verifier.rs
```diff
@@ -788,11 +796,13 @@ fn process_filter(
...
         tag_ @ "schema_id" => precess_filed(tag_, filter.schema_id.to_string(), tag_value),
-        tag_ @ "schema_issuer_did" => precess_filed(tag_, &filter.schema_issuer_did, tag_value),
+        tag_ @ "schema_issuer_did" => {
+            precess_filed(tag_, filter.schema_issuer_id.to_owned(), tag_value)
+        }
         tag_ @ "schema_name" => precess_filed(tag_, &filter.schema_name, tag_value),
         tag_ @ "schema_version" => precess_filed(tag_, &filter.schema_version, tag_value),
         tag_ @ "cred_def_id" => precess_filed(tag_, &filter.cred_def_id.to_string(), tag_value),
-        tag_ @ "issuer_did" => precess_filed(tag_, &filter.issuer_did, tag_value),
+        tag_ @ "issuer_did" => precess_filed(tag_, filter.issuer_id.to_owned(), tag_value),
```
TimoGlastra, line 805: I see you left the keys as `_did`. We need to think a bit about backwards interop here. We probably want `_did` to be valid, but only for legacy (unqualified identifiers). For all others we need to use the `_id` variant.

This way we keep the old exchange valid

TimoGlastra: OR we require the input to be `_id` and the user of the library must manually transform it from `_did` to `_id`. But I think it's maybe easier to support both in here?

anoncreds/src/data_types/anoncreds/cred_def.rs
blu3beri: yes, you are correct. I always do both, no clue why.

anoncreds/src/services/verifier.rs
blu3beri: Hmm yeah I am not too sure. I did not notice the key, so that is a good catch. 

Thinking whether this would work okay:

```rust
        tag_ @ "issuer_did" => precess_filed("issuer_id", filter.issuer_id.to_owned(), tag_value),
        tag_ @ "issuer_id" => precess_filed(tag_, filter.issuer_id.to_owned(), tag_value),

```

TimoGlastra: Hmm not sure if we should mix them. Where is this code used for? Is it only for internal processing? We don't want to modify the output model. And they probably shouldn't be allowed to be defined at the same time?

TimoGlastra: Well, I'd really like to make the old indy flow (pre standardization) still valid. If we remove the `_did` part we break interop. Until now, all breaking changes have been on the ledger models, which you resolve on your own. So if I support new AnonCreds, and you support old anoncreds we're all good (as long as I also use old).

TimoGlastra: Maybe something to discuss during the WG call. It has quite some impact on interop. Maybe we keep both in the spec, but we make the anoncreds lib only support, `_id`. But that wouldn't make a lot of sense also...

blu3beri: Another thing to note is we added validation now to the `issuer_id`. So if we were to reuse it with a legacy identifier, it would not work (as validation will fail).

TimoGlastra: Legacy indy identifiers will still be valid for dids, meaning the issuer_id won't be always an uri

blu3beri: Okay then I need to redo the validation for the identifiers as it can be a URI or the legacy did indy.

Would this suffice:

```
1. Check if URI
2. If URI, success
3. If not URI, check legacy indy identifier
  - Check length (not sure which lengths, I thought there were 2 different lengths)
  - Check if all chars are in the base58 alphabet
```

TimoGlastra: Yes. I think the length is 21 or 22 (we have a regex in AFJ)

blu3beri: Awesome. Will add that to this PR.

blu3beri: It should be fixed with the latest commit.

TimoGlastra:
@blu3beri on the last WG call we came to the conclusion that:
- Only legacy indy identifiers can use `_did` for backwards compatiblity
- All other identifier types (qualified indy, but also other methods) MUST use `_id`
- They can't be used at the same time

That means we need to support both in this library, and that we need to verify the above rules

blu3beri:
> * They can't be used at the same time

Do you mean they can't be used in the same proof request of for s specific group within the request?

TimoGlastra:
> Do you mean they can't be used in the same proof request of for s specific group within the request?

Hm. If you're using the new syntax anywhere in the proof request, you could just as well use it everywhere. But that's probalby more complex to verify?

TimoGlastra (approved):
Nice 👍

anoncreds/src/services/verifier.rs
```diff
@@ -788,11 +835,17 @@ fn process_filter(
     );
     match tag {
         tag_ @ "schema_id" => precess_filed(tag_, filter.schema_id.to_string(), tag_value),
-        tag_ @ "schema_issuer_did" => precess_filed(tag_, &filter.schema_issuer_did, tag_value),
+        tag_ @ "schema_issuer_did" => {
+            precess_filed(tag_, filter.schema_issuer_id.to_owned(), tag_value)
```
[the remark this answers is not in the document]
blu3beri: Yeah it just maps to the issuer_id in the schema. It got a bit confusing trying to find the answer, but below we get the issuer_id from the schema and map it to the `filter.schema_issuer_Id`. the `schema.issuer_id` can be legacy or new. So we check if the provided value for `schema_issuer_did` is legacy and is inside the `schema.issuer_id`.

```rust
    Ok(Filter {
        schema_id: schema_id.to_owned(),
        schema_name: schema.name.to_owned(),
        schema_version: schema.version.to_owned(),
        schema_issuer_id: schema.issuer_id.to_owned(),
        issuer_id: cred_def.issuer_id.to_owned(),
        cred_def_id: cred_def_id.to_owned(),
    })

```

PR #55, opened 2023-01-11
added schema attrNames validation

blu3beri:


anoncreds/src/services/issuer.rs
```diff
@@ -38,6 +38,8 @@ pub fn create_schema(
         attr_names
     );
 
+    attr_names.validate()?;
```
TimoGlastra, line 41: what are validating here?

TimoGlastra (approved):

anoncreds/src/services/issuer.rs
blu3beri: The length must be bigger than 0 and smaller than the max_count (125)

```rust
impl Validatable for AttributeNames {
    fn validate(&self) -> Result<(), ValidationError> {
        if self.0.is_empty() {
            return Err("Empty list of Schema attributes has been passed".into());
        }

        if self.0.len() > MAX_ATTRIBUTES_COUNT {
            return Err(format!(
                "The number of Schema attributes {} cannot be greater than {}",
                self.0.len(),
                MAX_ATTRIBUTES_COUNT
            )
            .into());
        }
        Ok(())
    }
}
```

PR #57, opened 2023-01-11
removed required prover_did

blu3beri:

### DIFF
--- a/anoncreds/src/data_types/mod.rs
+++ b/anoncreds/src/data_types/mod.rs
@@ -3,7 +3,7 @@ mod utils {
     pub use indy_utils::{qualifiable, Qualifiable};
 }
 
-pub use indy_utils::did;
+//pub use indy_utils::did;
 pub use indy_utils::keys;
 pub use indy_utils::{invalid, ConversionError, Validatable, ValidationError};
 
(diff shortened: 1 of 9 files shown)


anoncreds/src/data_types/mod.rs
```diff
@@ -3,7 +3,7 @@ mod utils {
     pub use indy_utils::{qualifiable, Qualifiable};
 }
 
-pub use indy_utils::did;
+//pub use indy_utils::did;
```
TimoGlastra, line 6: Commented

anoncreds/src/services/prover.rs
```diff
@@ -46,6 +48,10 @@ pub fn create_credential_request(
         credential_offer
     );
 
+    let rand_str = String::from_utf8(thread_rng().sample_iter(&Alphanumeric).take(22).collect())
```
TimoGlastra, line 51: The value prover_did should become optional in the whole credential request. If not present in the request, the issuer will generate it. 

So if no prover_did is present we don't include it in the request

blu3beri:
@TimoGlastra I think I fixed the issue, it was a bit unclear to me but this should hopefully be it.

TimoGlastra (approved):

PR #64, opened 2023-01-16
Revocable checks

### LINKED ISSUE #42: Support for mixed proof with revocable/non-revocable credentials
This is an issue with existing anoncreds implementations (indy-sdk and credx) and is illustrated in an aca-py integration test:

https://github.com/hyperledger/aries-cloudagent-python/blob/main/demo/features/0454-present-proof.feature#L120

When a proof includes both revocable and non-revocable credentials, and the request includes the revocation timestamp at the REQUEST level, then the proof will fail (even though it should pass).

Note that when the revocation interval is requested at the ATTRIBUTE level, the proof will pass (see integration test https://github.com/hyperledger/aries-cloudagent-python/blob/main/demo/features/0454-present-proof.feature#L100)

whalelephant:
Closes #42 and updates e2e test framework

### DIFF
--- a/anoncreds/src/data_types/anoncreds/rev_reg.rs
+++ b/anoncreds/src/data_types/anoncreds/rev_reg.rs
@@ -91,6 +91,7 @@ impl RevocationStatusList {
         revoked: Option<BTreeSet<u32>>,
         timestamp: Option<u64>,
     ) -> Result<(), error::Error> {
+        // only update if input is Some
         if let Some(reg) = registry {
             self.registry = Some(reg)
         };
@@ -120,6 +121,7 @@ impl RevocationStatusList {
                 *bit = true;
             }
         }
+        // only update if input is Some
         if let Some(t) = timestamp {
             self.timestamp = Some(t);
         }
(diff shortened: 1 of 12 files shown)


anoncreds/src/data_types/anoncreds/rev_reg.rs
```diff
@@ -73,6 +73,26 @@ impl RevocationStatusList {
         self.revocation_list.get(idx).as_deref().copied()
     }
 
+    pub fn revoke(&mut self, idx: usize) -> Option<bool> {
```
blu3beri, line 76: Might be good to add some code doc when it returns true / false / none.

anoncreds/tests/utils/mock.rs
```diff
@@ -0,0 +1,443 @@
...
+#[derive(Debug)]
+pub struct Mock<'a> {
+    pub issuer_wallets: HashMap<&'a str, IssuerWallet<'a>>,
+    pub prover_wallets: HashMap<&'a str, ProverWallet<'a>>,
+    pub ledger: Ledger<'a>,
+    pub tails_path: &'a str,
+    pub max_cred_num: u32,
+}
+
+impl<'a> Mock<'a> {
```
blu3beri, line 47: I have never really done any intensive mocking with Rust, but would it not be easier to pull in a dependency for this? I have no issue with keeping this, but we might have a bit less code with a mock dependency.

whalelephant:
Addressed by #102

PR #65, opened 2023-01-17
build(javascript): added initial pipeline

blu3beri:


TimoGlastra (approved):
Nice 👍

TimoGlastra:
> @TimoGlastra Can you enable the `build.yml` as a required workflow before merging? I removed the `rusttest.yml` in this pr so we have no listing and stuff done when we want to create a PR.
> 
> 
> 
> Or should Ry do that?

I think you have the same permissions as me! 

I've enabled the following checks: is that good?

image

blu3beri:
> > @TimoGlastra Can you enable the `build.yml` as a required workflow before merging? I removed the `rusttest.yml` in this pr so we have no listing and stuff done when we want to create a PR.
> > Or should Ry do that?
> 
> I think you have the same permissions as me!
> 
> I've enabled the following checks: is that good?
> 
> image

Huh that's weird. It shows the Test suite but I accidentally removed that. So it might only pick up actions from `main`.

Also, I can't seem to change the permissions, maybe you have some additional access within Hyperledger?

TimoGlastra:
Ok, you can merge when you're ready and I'll update it in settings afterwards

PR #70, opened 2023-01-19
Tests: multiple credentials

whalelephant:
Note: the previous `CredentialRevocationConfig` struct used `RevocationRegistry` and had to be removed in this PR.

### DIFF
--- a/anoncreds/src/services/issuer.rs
+++ b/anoncreds/src/services/issuer.rs
@@ -1,6 +1,7 @@
 use rand::distributions::Alphanumeric;
 use rand::{thread_rng, Rng};
 use std::collections::{BTreeSet, HashSet};
+use std::convert::{From, Into};
 use std::iter::FromIterator;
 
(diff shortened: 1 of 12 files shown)


anoncreds/src/services/issuer.rs
```diff
@@ -219,61 +218,60 @@ pub fn create_revocation_status_list(
     )
 }
 
+/// Update the timestamp only without changing any actual state
+pub fn update_revocation_status_list_timestamp_only(
```
blu3beri, line 222: This function is not used anywhere right? Maybe we should add some tests for it?

blu3beri (approved):
One comment about updating the status list timestamp method, otherwise it looks good to me :D

anoncreds/src/services/issuer.rs
whalelephant: you are right, I will update the issue of the Todos for revocation status list

PR #84, opened 2023-01-26
fix(js): added release step to the pipeline

berendsliedrecht:


TimoGlastra (commented):
It seems this is missing the updates to the build file?

blu3beri:
> It seems this is missing the updates to the build file?

Added it now. I guess I restored some work and accidentally included the addition of the release stuff.

TimoGlastra (approved):
Will it work that one is called `darwin` and the other is called `macos` in the release? Will node-pre-gyp also make that differences in the `platform`? 

no, good catch ;)

PR #87, opened 2023-01-28
fix(js): missing export and return type

genaris:
Some minor fixes in JavaScript wrapper:

- Remove the need of passing a `Presentation` instance in `verify`, as it is an instance method and can get the handle by itself
- Expose `RevocationStatusList`
- Fix `RevocationRegistryDefinition.load` method


TimoGlastra (approved):
🙌

PR #91, opened 2023-02-01
fix react native

blu3beri:
Native side of the wrapper should be good but for some reason the native module does not register. I will see if I can fix this somehow, but it might be that it works when its installed via NPM.


blu3beri:
@TimoGlastra somehow the linking does not seem to work when I manually try to do it (android and iOS). I was thinking maybe making a dev release and trying that to see would be worth it. 

Could you give this a review? The js side is still incomplete but the cpp side should be good :).

TimoGlastra (approved):

PR #95, opened 2023-02-02
build: build xc frameworks with dylibs

blu3beri:
- Build the xcframework with dynamic instead of static libraries
- used in the CI now

### Notes

- Dynamic libraries and frameworks pushing to the App Store can fail if it is not setup correctly I think.
  - Since I am not an iOS developer I "botched" together some of the resources about constructing a `.framework` online, but we have to push an app to the app store to see whether it is done properly. if anyone knows anyone with experience like this, help would be awesome!
- This is by no means a perfect shell script. It can be refactored later, but it does the job for now.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -381,19 +381,15 @@ jobs:
       - name: Fetch static libraries
         uses: actions/download-artifact@v3
       - run: >
-          lipo -create aarch64-apple-ios-sim/libanoncreds.a \
-                       x86_64-apple-ios/libanoncreds.a \
-               -output libanoncreds.a
-      - run: >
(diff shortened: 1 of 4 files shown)


TimoGlastra (approved):
Interesting. Why is this different from the Askar and Indy VDR repos?

blu3beri:
> Interesting. Why is this different from the Askar and Indy VDR repos?

This will be done for those as well. I got started on this as the static libraries did not expose symbols (non-deterministically of course) and it made the iOS build fail.

If this CI works and it is released with the correct framework which will work in a new project, I will add the same scripts to askar and ver for sure! (no more bundle of 150MiBs as well :D)

PR #100, opened 2023-02-05
fix: parameters when creating credentials without revocation

genaris:
When attempting to create some credentials without any kind of revocation configuration, under JS I've noticed that it was not properly handling some optional parameters for creating and processing credentials:
- In `create_credential`, `revocationConfiguration` was sending a default CredRevInfoStruct, that library was not interpreting as null so it was throwing an error about missing tails file path
- Also in `create_credential`, JS wrapper was not accepting an empty  `revocationStatusList`. In addition, once solved, Rust part was throwing an 'Invalid object handle' error when trying to load the object as it was mandatory
- In `process_credential`, JS wrapper was not accepting an empty `revocationRegistryDefinition`

I'm not sure if the solutions implemented in this PR are completely correct, even if they seem to work fine. So please @whalelephant and @blu3beri (as the authors of this revocation stuff in Rust and FFI wrapper in JS) let me know about any suggestion you may have (or feel free to add some commits to this PR or even completely disregard it if it's too bad! :stuck_out_tongue: )

### DIFF
--- a/src/ffi/credential.rs
+++ b/src/ffi/credential.rs
@@ -125,7 +125,10 @@ pub extern "C" fn anoncreds_create_credential(
             cred_request.load()?.cast_ref()?,
             cred_values.into(),
             rev_reg_id,
-            rev_status_list.load()?.cast_ref().ok(),
+            rev_status_list.opt_load()?
+                .as_ref()
+                .map(AnonCredsObject::cast_ref)
+                .transpose()?,
             revocation_config
                 .as_ref()
                 .map(RevocationConfig::as_ref_config)
--- a/src/ffi/credential.rs
+++ b/src/ffi/credential.rs
@@ -125,7 +125,8 @@ pub extern "C" fn anoncreds_create_credential(
             cred_request.load()?.cast_ref()?,
             cred_values.into(),
             rev_reg_id,
-            rev_status_list.opt_load()?
+            rev_status_list
+                .opt_load()?
                 .as_ref()
                 .map(AnonCredsObject::cast_ref)
                 .transpose()?,
--- a/wrappers/javascript/anoncreds-nodejs/src/NodeJSAnoncreds.ts
+++ b/wrappers/javascript/anoncreds-nodejs/src/NodeJSAnoncreds.ts
@@ -285,7 +285,7 @@ export class NodeJSAnoncreds implements Anoncreds {
 
     const credentialEntries = options.credentials.map((value) => {
       const { credential, timestamp, revocationState: rev_state } = serializeArguments(value)
-      return CredentialEntryStruct({ credential, timestamp, rev_state })
+      return CredentialEntryStruct({ credential, timestamp: timestamp ?? -1, rev_state: rev_state ?? 0 })
     })
 
     const credentialEntryList = CredentialEntryListStruct({
--- a/wrappers/javascript/anoncreds-nodejs/src/NodeJSAnoncreds.ts
+++ b/wrappers/javascript/anoncreds-nodejs/src/NodeJSAnoncreds.ts
@@ -125,7 +125,6 @@ export class NodeJSAnoncreds implements Anoncreds {
       credentialOffer,
       credentialRequest,
       revocationRegistryId,
-      revocationStatusList,
     } = serializeArguments(options)
 
     const attributeNames = StringListStruct({
@@ -172,8 +171,8 @@ export class NodeJSAnoncreds implements Anoncreds {
       attributeRawValues as unknown as Buffer,
       attributeEncodedValues as unknown as Buffer,
       revocationRegistryId,
-      revocationStatusList ?? 0,
-      revocationConfiguration ? revocationConfiguration.ref().address() : 0,
+      options.revocationStatusList?.handle ?? 0,
+      revocationConfiguration?.ref().address() ?? 0,
       credentialPtr
     )
     handleError()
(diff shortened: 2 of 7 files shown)


TimoGlastra (approved):
LGTM, but I'm not sure on the optional loading in Rust / the passing of `0`,  we need the rust pros for this 🦀

src/ffi/credential.rs
```diff
@@ -125,7 +125,11 @@ pub extern "C" fn anoncreds_create_credential(
             cred_request.load()?.cast_ref()?,
             cred_values.into(),
             rev_reg_id,
-            rev_status_list.load()?.cast_ref().ok(),
+            rev_status_list
+                .opt_load()?
+                .as_ref()
+                .map(AnonCredsObject::cast_ref)
+                .transpose()?,
```
blu3beri, line 132:
```suggestion
            rev_status_list
                .opt_load()?
                .cast_ref()
                .transpose()?,
```

Would this not work?

wrappers/javascript/anoncreds-nodejs/src/NodeJSAnoncreds.ts
```diff
@@ -172,8 +172,8 @@ export class NodeJSAnoncreds implements Anoncreds {
       attributeRawValues as unknown as Buffer,
       attributeEncodedValues as unknown as Buffer,
       revocationRegistryId,
-      revocationStatusList,
-      revocationConfiguration.ref(),
+      revocationStatusList ?? 0,
```
blu3beri, line 175: I think we just have to pass `undefined` for this and not `0`.

wrappers/javascript/anoncreds-nodejs/src/NodeJSAnoncreds.ts
```diff
@@ -172,8 +172,8 @@ export class NodeJSAnoncreds implements Anoncreds {
...
+      revocationConfiguration ? revocationConfiguration.ref().address() : 0,
```
blu3beri, line 176: Same for here. I think just the value / undefined works.

wrappers/javascript/anoncreds-nodejs/src/library/bindings.ts
```diff
@@ -35,7 +35,7 @@ export const nativeBindings = {
       StringListStruct,
       FFI_STRING,
       FFI_OBJECT_HANDLE,
-      refType(CredRevInfoStruct),
+      FFI_OBJECT_HANDLE,
```
blu3beri, line 38: GH says some values are not used in this file anymore. Can you remove them?

blu3beri (dismissed):

wrappers/javascript/anoncreds-nodejs/src/NodeJSAnoncreds.ts
genaris: The problem is it's expecting a number, so if I send undefined I get

```TypeError: error setting argument 9 - writeUInt64: Number/String 64-bit value required```

Would it be more elegant to change the code to something like

```ts
      revocationStatusList ?? new ObjectHandle(0).handle,
```
?

wrappers/javascript/anoncreds-nodejs/src/NodeJSAnoncreds.ts
genaris: In this case it's at compilation time where I get the `number | undefined is not assignable to number`. Maybe there is a way for refType to allow it?

src/ffi/credential.rs
genaris: If I do it directly it gives me a compilation error:

```
error[E0599]: no method named `cast_ref` found for enum `std::option::Option` in the current scope
```

genaris:
@blu3beri just to make a bigger controversy, I've just added some fixes for optional parameters in `CredentialEntryStruct`: 

- timestamp: I had to default it to -1 because otherwise anoncreds would take it as valid
- rev_state: Defaulted to 0 as other cases in this PR

So basically the question is how can we define the structures in order to accept these values, or how to find an elegant way to define them.

src/ffi/credential.rs
blu3beri: Okay yeah that makes sense. I think your implementation is good.

wrappers/javascript/anoncreds-nodejs/src/NodeJSAnoncreds.ts
```diff
@@ -172,8 +171,8 @@ export class NodeJSAnoncreds implements Anoncreds {
       attributeRawValues as unknown as Buffer,
       attributeEncodedValues as unknown as Buffer,
       revocationRegistryId,
-      revocationStatusList,
-      revocationConfiguration.ref(),
+      options.revocationStatusList?.handle ?? 0,
```
blu3beri, line 174: Isn't this handled in the `serializeArguments` function? I would strongly prefer not to send the `0` as that can be a valid handle number. 

If `serializeArguments` does not handle this we should 100% do it there. Maybe we should just use `USIZE::MAX`?

blu3beri (commented):
Just one comment regarding the use of `0` for an unexacting handle type.

wrappers/javascript/anoncreds-nodejs/src/NodeJSAnoncreds.ts
genaris: The problem with `serializeArguments` is that it receives the object as undefined so it does not know which type of object it is supposed to be (i.e. typeof arg = 'undefined'). So it returns NULL, which works fine for optional strings but not for optional object handles (whose type is `size_t`).

TimoGlastra (approved):

PR #102, opened 2023-02-09
Support combination of revocable / non-revocable credentials and timestamp override

whalelephant:


TimoGlastra (approved):
Nice. Like the extensive comments!

src/ffi/presentation.rs
```diff
@@ -195,6 +195,50 @@ pub extern "C" fn anoncreds_create_presentation(
...
+///
+/// A `nonrevoked_interval = {from: 50, to: 150}` should accept both List0 and
+/// List100.  
+///
+#[derive(Debug)]
+#[repr(C)]
+pub struct FfiNonrevokedIntervalOverride<'a> {
+    rev_reg_def_id: FfiStr<'a>,
+    /// Timestamp in the `PresentationRequest`
+    requested_from_ts: i64,
```
blu3beri, line 221: Might be good later on to use i32s here. I think we map them to a u64 later and it would help in the JS side where every number is an f32.

src/services/helpers.rs
```diff
@@ -134,55 +135,108 @@ pub fn build_sub_proof_request(
...
-mod tests {
-    use super::*;
+pub fn get_revealed_attributes_for_credential(
+    sub_proof_index: usize,
+    requested_proof: &RequestedProof,
+    pres_req: &PresentationRequestPayload,
+) -> Result<(Vec<AttributeInfo>, Option<NonRevokedInterval>)> {
+    trace!("_get_revealed_attributes_for_credential >>> sub_proof_index: {:?}, requested_credentials: {:?}, pres_req: {:?}",
+           sub_proof_index, requested_proof, pres_req);
+    let mut nonrevoked_interval: Option<NonRevokedInterval> = None;
```
blu3beri, line 149:
```suggestion
    let mut non_revoked_interval: Option<NonRevokedInterval> = None;
```

blu3beri (approved):

src/ffi/presentation.rs
whalelephant: ah good idea! fixed

src/services/helpers.rs
whalelephant: fixed

TimoGlastra:
@whalelephant is this ready to merge? Any blockers here?

TimoGlastra:
Do we need any updates to the FFI wrappers? If so could you point me in the right direction on what needs to be fixed?

whalelephant:
The wrapper has been implemented and this is ready to be merged. But it is breaking change for `anoncreds_verify_presentation` interface.

blu3beri:
Can we merge this or would another review be good here?

PR #104, opened 2023-02-10
chore: update version to 0.1.0-dev.5

TimoGlastra:
Signed-off-by: Timo Glastra <timo@animo.id>

### DIFF
--- a/build-xcframework.sh
+++ b/build-xcframework.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env sh
 
 NAME="anoncreds"
-VERSION="0.1.0-dev.4"
+VERSION="0.1.0-dev.5"
 BUNDLE_IDENTIFIER="org.hyperledger.$NAME"
 LIBRARY_NAME="lib$NAME.dylib"
 XC_FRAMEWORK_NAME="$NAME.xcframework"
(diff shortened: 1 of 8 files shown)


build-xcframework.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env sh
 
 NAME="anoncreds"
-VERSION="0.1.0-dev.4"
+VERSION="0.1.0-dev.5"
```
blu3beri, line 4: I will do the same fix here as I did for askar. We will not have to update this ourselves anymore.

wrappers/javascript/anoncreds-nodejs/test/bindings.test.ts
```diff
@@ -8,7 +8,7 @@ describe('bindings', () => {
   test('version', () => {
     const version = anoncreds.version()
 
-    expect(version).toEqual('0.1.0-dev.4')
+    expect(version).toEqual('0.1.0-dev.5')
```
blu3beri, line 11: This should probably be removed later on :).

blu3beri (approved):

wrappers/javascript/anoncreds-nodejs/test/bindings.test.ts
genaris: Or changed to match a regex

PR #119, opened 2023-02-27
fix: issuer id in rev reg def must be the same as the credential def

blu3beri:


src/services/issuer.rs
```diff
@@ -139,6 +139,12 @@ where
     let cred_def_id = cred_def_id.try_into()?;
     let issuer_id = issuer_id.try_into()?;
 
+    if issuer_id != cred_def.issuer_id {
```
TimoGlastra, line 142: I think we also need to add this check when working with objects that contain a reference to the credential definition. If the revocation registry definition is created outside of this lib, it could be that we get a rev reg def for a cred def that is from a different issuer id.

In addition we also need to make sure we do the same check for revocation registry and revocation status lists.

blu3beri: Let me see, I might be able to add this in the validate function.

blu3beri: I don't think it is actually an issue. `create_revocation_registry` is the public API of this library (ignoring FFI). In this method we check it. The struct could be instantiated outside of the library, but since the input does not include the `cred_def`, we cannot validate whether the ids are correct. We would need to change quite a bit about the internals for this to work and I also think that this check cannot be relied upon (I see it more as a nice-to-have).

tl;dr I will add the checks to rev reg and rev status list, but I won't add it in another place as we do not have the correct data there.

blu3beri: Another note, the rev status list does not take an `issuer_id` and we don't have a revocation registry anymore right? I am a bit lost.

TimoGlastra: The revocation status list has the same issuerId as the revocation registry definition. However it may be odd that only the status list doesn't have an issuer id ... 🤔

blu3beri:
@TimoGlastra it is ready for a re-review again

wrappers/javascript/anoncreds-nodejs/src/NodeJSAnoncreds.ts
```diff
@@ -405,6 +430,7 @@ export class NodeJSAnoncreds implements Anoncreds {
       revocationRegistryDefinitions,
       revocationRegistryDefinitionIds,
       revocationStatusLists,
+      nonRevokedIntervalOverrideList as unknown as Buffer,
```
TimoGlastra, line 433: do you think we can improve all the casts etc.. a bit?

TimoGlastra (approved):

wrappers/javascript/anoncreds-nodejs/src/NodeJSAnoncreds.ts
blu3beri: Yeah it annoys me a lot... It comes down to ref-napi and struct-napi not being type-compatible (everything does work though...) and struct-napi having very complex types.

We do have to cast, but we might be able to extract some of that to a function and reuse it. 

So, we would have like `createNativeNonRevokedIntervalOverrideList(...) -> Buffer` which does the casting only there.

PR #120, opened 2023-02-28
fix: python wrapper type hint tweaks

dbluhm:
This PR refines type hints in the python wrapper. These corrections make static type checkers like pyright/pylance happier.

I'll open this as a draft for now and add more commits as I run into type complaints from pyright.


blu3beri:
@dbluhm should we merge this PR separately or are you going to incorporate this into #145

dbluhm:
I'll combine these changes with #145

PR #121, opened 2023-02-28
add required entropy instead of optional prover_did

blu3beri:

### DIFF
--- a/src/data_types/cred_request.rs
+++ b/src/data_types/cred_request.rs
@@ -27,7 +27,7 @@ impl Validatable for CredentialRequest {
                 } else {
                     Ok(())
                 }
-            },
+            }
             None => {
                 if self.cred_def_id.is_legacy() {
                     if let Some(prover_did) = self.prover_did.clone() {
@@ -89,10 +89,19 @@ pub struct CredentialRequestMetadata {
 
 impl Validatable for CredentialRequestMetadata {}
 
-
 #[cfg(test)]
 mod cred_req_tests {
-    use crate::{prover::create_credential_request, data_types::{schema::AttributeNames, cred_def::{SignatureType, CredentialDefinition, CredentialKeyCorrectnessProof}, master_secret::MasterSecret, cred_offer::CredentialOffer}, issuer::{create_schema, create_credential_definition, create_credential_offer}, types::CredentialDefinitionConfig};
+    use crate::{
+        data_types::{
+            cred_def::{CredentialDefinition, CredentialKeyCorrectnessProof, SignatureType},
+            cred_offer::CredentialOffer,
+            master_secret::MasterSecret,
+            schema::AttributeNames,
+        },
+        issuer::{create_credential_definition, create_credential_offer, create_schema},
+        prover::create_credential_request,
+        types::CredentialDefinitionConfig,
+    };
 
     use super::*;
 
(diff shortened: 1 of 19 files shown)


src/data_types/cred_request.rs
```diff
@@ -5,8 +5,7 @@ use super::{cred_def::CredentialDefinitionId, nonce::Nonce};
 
 #[derive(Debug, Deserialize, Serialize)]
 pub struct CredentialRequest {
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub prover_did: Option<String>,
+    pub entropy: String,
```
TimoGlastra, line 8: entropy is only required if not using legacy indy identifiers to keep backwards compatibility

TimoGlastra: Maybe we can also make it optional when creating a request and generate something otherwise.

So:
- if cred_def_id is not legacy and  `entropy` is provided, use that
- if cred_def_id is not legacy and no `entropy` is provided, generate an entropy value
- if cred_def_id is legacy and no `entropy` is provided generate a `prover_did` value

Not sure if this is the best approach?

blu3beri: I assumed it was required due to this comment: https://github.com/hyperledger/anoncreds-spec/issues/107#issuecomment-1419243041, although I did forget the discussion in the WG around this so you might be correct.

We would basically have `entropy` required with new identifiers and `prover_did` with legacy identifiers, correct? 

I think I can do some nice `serde` stuff here to make that work with a single value. 

Re the approach:

I think the first one is incorrect, as entropy is required (due to https://github.com/hyperledger/anoncreds-spec/issues/107#issuecomment-1419243041).

Same goes for point two, entropy is required so we would fail if its not supplied.

For point three, `prover_did` used to be required as well. This would mean that when legacy identifiers are used, we should change the key of `entropy` to `prover_did` in the model (and validate its a correct `prover_did`). 

Does that work for you?

TimoGlastra: > and prover_did with legacy identifiers, correct?

legacy identifiers also support `entropy`. So it's mostly to keep backwards compat

> Does that work for you?

Yes. It would prevent `entropy` from being used when using legacy identifiers though

blu3beri: Ahhh okay. So we can also do that `prover_did` would never be supplied. So if some entropy is supplied we will use the key `entropy`, correct?

TimoGlastra: Yeah, just thinking out loud here. It may be messy to have multiple input parameters related to this, while in 99% of cases you'd want to generate it anyway.

But it may be funky if you don't understand the behaviour

blu3beri: My current plan:

- Entropy is a required input
- If using new identifiers, use the provided entropy
- if using old identifiers
   - entropy is compatible with `prover_did` requirements: use `prover_did`
   - entropy is incompatible with `prover_did` requirements: use `entropy`

It is a bit magic / unknown behaviour, but it might be the best we can do.

blu3beri: Provide both entropy and prover_did and prover_did only when using legacy identifiers.

TimoGlastra (approved):
Perfect!

PR #128, opened 2023-03-03
Add linux arm64 as release target

lauravuo:
Add linux arm64 distribution as target and a step for preparing the build environment for the aarch64-tooling.


TimoGlastra (approved):
Nice!

blu3beri:
Can't we use cross for this to be more in-line with the other packages? Otherwise I am okay with it.

lauravuo:
Closing, as it seems the support is coming with another track.

PR #129, opened 2023-03-03
refactor(js)!: use JSON objects in API instead of strings and instances

genaris:
This is an attempt (without doing anything revolutionary) to make JS API a bit more simple to use, avoiding the so-many conversions from JSON to string  that we found during integration with AFJ. 

For most methods that are currently receiving `AnoncredsObject` instances, now we accept also passing JSON objects. This makes the caller to not care about object instantiation and clearing handles, which is done internally. Hopefully, in a normal flow the caller will need only to care about freeing a few objects (like Credentials).

I'm not sure if I'm using the `try-finally` method correctly (I'm particularly worried about the order of execution of finally if an exception is thrown) so any comments on this approach would be appreciated. I wanted to make this process of "checking object type and loading from JSON/add to handles to clear array" more elegant, but I think I'll need some more TypeScript lessons from @TimoGlastra and @blu3beri before attempting to do that 😛 .

Some tests were added to this repo, but I didn't check this as a patch of AFJ to see how is it going, something I'd like to do before merging. 

Summary of changes: 

- Refactor `load` in AnonCredsObjects to `fromJson`, loading from a JSON object rather than a string
- Refactor `toJson` in AnoncredsObject to return a JSON object rather than a string
- All constructors and methods in API AnoncredsObjects accept both AnonCredsObjects and JSON objects. This is the most tricky part: for every passed JSON object, an  ephemeral AnonCredsObject instance will be created and must be cleared before returning to the caller
- Add some missing fromJson (like the one from Credential)
- Rename keyProof -> keyCorrectnessProof to match the naming of other fields

### DIFF
--- a/wrappers/javascript/anoncreds-nodejs/test/api.test.ts
+++ b/wrappers/javascript/anoncreds-nodejs/test/api.test.ts
@@ -175,7 +175,7 @@ describe('API', () => {
 
     expect(presentation.handle.handle).toStrictEqual(expect.any(Number))
 
-    const verify = Presentation.fromJson(presentation.toJson()).verify({
+    const verify = presentation.verify({
       presentationRequest,
       schemas: { ['mock:uri']: schema },
       credentialDefinitions: { ['mock:uri']: credentialDefinition },
@@ -319,7 +319,7 @@ describe('API', () => {
 
     expect(presentation.handle.handle).toStrictEqual(expect.any(Number))
 
-    const verify = Presentation.fromJson(presentation.toJson()).verify({
+    const verify = presentation.verify({
       presentationRequest,
       schemas: { ['mock:uri']: schema },
       credentialDefinitions: { ['mock:uri']: credentialDefinition },
--- a/wrappers/javascript/anoncreds-nodejs/test/api.test.ts
+++ b/wrappers/javascript/anoncreds-nodejs/test/api.test.ts
@@ -368,6 +368,7 @@ test('create and verify presentation passing only JSON objects as parameters)', 
   const masterSecretId = 'master secret id'
 
   const { credentialRequestMetadata, credentialRequest } = CredentialRequest.create({
+    entropy: 'entropy',
     credentialDefinition: credentialDefinition.toJson(),
     masterSecret: masterSecret.toJson(),
     masterSecretId,
--- a/wrappers/javascript/anoncreds-nodejs/test/api.test.ts
+++ b/wrappers/javascript/anoncreds-nodejs/test/api.test.ts
@@ -22,32 +22,30 @@ describe('API', () => {
   test('create and verify presentation', () => {
     const nonce = anoncreds.generateNonce()
 
-    const presentationRequest = PresentationRequest.load(
-      JSON.stringify({
-        nonce,
-        name: 'pres_req_1',
-        version: '0.1',
-        requested_attributes: {
-          attr1_referent: {
-            name: 'name',
-            issuer: 'mock:uri',
-          },
-          attr2_referent: {
-            name: 'sex',
-          },
-          attr3_referent: {
-            name: 'phone',
-          },
-          attr4_referent: {
-            names: ['name', 'height'],
-          },
+    const presentationRequest = PresentationRequest.fromJson({
+      nonce,
+      name: 'pres_req_1',
+      version: '0.1',
+      requested_attributes: {
+        attr1_referent: {
+          name: 'name',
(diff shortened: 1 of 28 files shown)


wrappers/javascript/anoncreds-shared/src/api/Credential.ts
```diff
@@ -1,66 +1,138 @@
...
 
 export type ProcessCredentialOptions = {
-  credentialRequestMetadata: CredentialRequestMetadata
-  masterSecret: MasterSecret
-  credentialDefinition: CredentialDefinition
-  revocationRegistryDefinition?: RevocationRegistryDefinition
+  credentialRequestMetadata: CredentialRequestMetadata | JsonObject
+  masterSecret: MasterSecret | JsonObject
+  credentialDefinition: CredentialDefinition | JsonObject
+  revocationRegistryDefinition?: RevocationRegistryDefinition | JsonObject
```
TimoGlastra, line 34: The usage of `JsonObject` means we don't have any typing requirements for the object. Should it be `CredentialDefinition` or `CredentialDefinitionJson` where `CredentialDefinitionJson` specifies the structure the json needs to have?

Fine to not do that for now, but it feels like we're leaving out the most useful thing a wrapper can provide, which is a clear interface

wrappers/javascript/anoncreds-shared/src/api/Credential.ts
```diff
@@ -1,66 +1,138 @@
...
-      revocationStatusList: options.revocationStatusList?.handle,
-    })
+    let credential
+    // Objects created within this method must be freed up
+    const objectHandles: ObjectHandle[] = []
+    try {
+      const credentialDefinition =
+        options.credentialDefinition instanceof CredentialDefinition
+          ? options.credentialDefinition.handle
+          : pushToArray(CredentialDefinition.fromJson(options.credentialDefinition).handle, objectHandles)
```
TimoGlastra, line 46: I think we can generically abstract this into a function:

```ts
const credentialDefinition = instanceOrCreateEphemeral(CredentialDefinition, options.credentialDefinition, objectHandles)
```

TimoGlastra (approved):
I think this is a great improvement to the public api of this library. Would like to have improvements to the interfaces for specific objects instead of `Record<string, unknown`, but that's also something we can add in additional PRs

wrappers/javascript/anoncreds-shared/src/api/Credential.ts
```diff
@@ -1,66 +1,138 @@
...
+        credentialOffer,
+        credentialRequest,
+        attributeRawValues: options.attributeRawValues,
+        attributeEncodedValues: options.attributeEncodedValues,
+        revocationRegistryId: options.revocationRegistryId,
+        revocationConfiguration: options.revocationConfiguration?.native,
+        revocationStatusList,
+      })
+    } finally {
+      objectHandles.forEach((handle) => handle.clear())
```
blu3beri, line 82: oh the clearing, ugh haha. 

I am wondering if in the future we can add a feature which removes the need for handles. This means that there might be some performance hits as we only work with raw values, but it might be good for the JS wrapper. 

Something for the future though, this is fine for now ;).

blu3beri (approved):

wrappers/javascript/anoncreds-shared/src/api/Credential.ts
genaris: I agree. It would be more or less the same model we have in `@aries-framework/anoncreds`, right? IIRC there was still some work to do in the typing for some objects, so probably will take a bit more time to get it 100% accurate.

wrappers/javascript/anoncreds-nodejs/test/api.test.ts
```diff
@@ -176,7 +174,7 @@ describe('API', () => {
 
     expect(presentation.handle.handle).toStrictEqual(expect.any(Number))
 
-    const verify = Presentation.load(presentation.toJson()).verify({
+    const verify = Presentation.fromJson(presentation.toJson()).verify({
```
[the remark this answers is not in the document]
genaris: In this case I think we can simply do `presentation.verify()` as we already have an instance of `Presentation`. I think the test previously did that in order to add additional checks of API methods. Now I've added an specific test for the JSON objects I think it's not needed to do that tricky conversion.

blu3beri:
@genaris can I merge this?

genaris:
> @genaris can I merge this?

Yes!

PR #130, opened 2023-03-05
fix(js): missing getXXXFromJson methods and several typos

genaris:
Adding the missing methods from react-native wrappers and making lots of minor fixes, mostly related to typos and outdated variable/method names.

At the moment working for a full issuance / present proof / verify proof flow without revocation (yes, also possible to issue a credential in React Native 🎉). However there is a memory leak leading to custom app crashes and variable corruption (!) when creating presentation.  Based on some debugging I did, the corrupted variables are strings in some FFI lists: it seems that strings are cleared or updated by the runtime before they are used by native library. Not sure if this makes sense, but I noticed that if I use short strings for fields like `issuerId`, `schemaId` and `credentialDefinitionId` (e.g. _'mock:uri'_) the flow works fine consistently, while if I use relatively long strings (real full DIDs, 50 characters long or so, no matter if it's only one of these variables), the flow crashes almost always. 🤔 

Also I think in some methods (like the aforementioned `createPresentation`) we are missing data deallocation but not completely sure how to handle that yet (@blu3beri you have all lottery numbers for this prize!! 😄). Work done in #129 will certainly help but I don't think if it's enough in a long-term app session. 

PS: All getXXXFromJson methods can be heavily refactored as they do almost the same thing. For the moment just wanted to be conservative to be focused in solving all issues, but any suggestion to make it cleaner will be appreciated!

### DIFF
--- a/wrappers/javascript/anoncreds-react-native/cpp/anoncreds.cpp
+++ b/wrappers/javascript/anoncreds-react-native/cpp/anoncreds.cpp
@@ -119,6 +119,293 @@ jsi::Value createCredentialDefinition(jsi::Runtime &rt, jsi::Object options) {
   return object;
 };
 
+// ===== AnonCreds Objects from JSON =====
+
+jsi::Value revocationRegistryDefinitionFromJson(jsi::Runtime &rt, jsi::Object options) {
+  auto json = jsiToValue<std::string>(rt, options, "json");
+
+  ObjectHandle revocationRegistryDefinitionP;
+
+    ByteBuffer b;
+
+    std::vector<uint8_t> jsonVector(json.begin(), json.end());
+    uint8_t *p = &jsonVector[0];
+    
+    b.data = p;
+    b.len = json.length();
+    
+  ErrorCode code = anoncreds_revocation_registry_definition_from_json(b, &revocationRegistryDefinitionP);
+  handleError(rt, code);
+
+  return jsi::Value(int(revocationRegistryDefinitionP));
+};
+
+jsi::Value revocationRegistryFromJson(jsi::Runtime &rt, jsi::Object options) {
+  auto json = jsiToValue<std::string>(rt, options, "json");
+
+  ObjectHandle revocationRegistryP;
+
+    ByteBuffer b;
+
+    std::vector<uint8_t> jsonVector(json.begin(), json.end());
+    uint8_t *p = &jsonVector[0];
+    
+    b.data = p;
+    b.len = json.length();
+    
+  ErrorCode code = anoncreds_revocation_registry_from_json(b, &revocationRegistryP);
+  handleError(rt, code);
+
+  return jsi::Value(int(revocationRegistryP));
+};
+
+jsi::Value presentationFromJson(jsi::Runtime &rt, jsi::Object options) {
+  auto json = jsiToValue<std::string>(rt, options, "json");
+
+  ObjectHandle presentationP;
+
+    ByteBuffer b;
+
+    std::vector<uint8_t> jsonVector(json.begin(), json.end());
+    uint8_t *p = &jsonVector[0];
+    
+    b.data = p;
+    b.len = json.length();
+    
+  ErrorCode code = anoncreds_presentation_from_json(b, &presentationP);
+  handleError(rt, code);
+
+  return jsi::Value(int(presentationP));
+};
+
+jsi::Value credentialOfferFromJson(jsi::Runtime &rt, jsi::Object options) {
+  auto json = jsiToValue<std::string>(rt, options, "json");
+
+  ObjectHandle credentialOfferP;
+
+    ByteBuffer b;
+
+    std::vector<uint8_t> jsonVector(json.begin(), json.end());
+    uint8_t *p = &jsonVector[0];
+    
+    b.data = p;
+    b.len = json.length();
+    
+  ErrorCode code = anoncreds_credential_offer_from_json(b, &credentialOfferP);
+  handleError(rt, code);
+
+  return jsi::Value(int(credentialOfferP));
+};
+
+jsi::Value schemaFromJson(jsi::Runtime &rt, jsi::Object options) {
+  auto json = jsiToValue<std::string>(rt, options, "json");
+
+  ObjectHandle schemaP;
+
+    ByteBuffer b;
+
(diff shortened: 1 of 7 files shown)


wrappers/javascript/anoncreds-react-native/cpp/anoncreds.cpp
```diff
@@ -103,22 +103,242 @@ jsi::Value createCredentialDefinition(jsi::Runtime &rt, jsi::Object options) {
...
+
+  ByteBuffer b = stringToByteBuffer(json);
+
+  ErrorCode code = anoncreds_revocation_state_from_json(b, &revocationStateP);
+  handleError(rt, code);
+
+  return jsi::Value(int(revocationStateP));
+};
+
+jsi::Value revocationRegistryDeltaFromJson(jsi::Runtime &rt, jsi::Object options) {
```
TimoGlastra, line 290: We don't have deltas anymore in the public API right?

wrappers/javascript/anoncreds-react-native/cpp/turboModuleUtility.cpp
```diff
@@ -160,9 +163,9 @@ FfiCredentialEntry jsiToValue(jsi::Runtime &rt, jsi::Object &options,
   if (value.isObject()) {
     jsi::Object valueAsObject = value.asObject(rt);
     auto credential = jsiToValue<ObjectHandle>(rt, valueAsObject, "credential");
-    auto timestamp = jsiToValue<int64_t>(rt, valueAsObject, "timestamp");
+    auto timestamp = jsiToValue<int64_t>(rt, valueAsObject, "timestamp", true);
     auto revocationState =
-        jsiToValue<ObjectHandle>(rt, valueAsObject, "revocationState");
+        jsiToValue<ObjectHandle>(rt, valueAsObject, "rev_state", true);
```
TimoGlastra, line 168: Why are we suddenly using snake_case here? Is that because  we're mapping to what the anoncreds binary expects?

wrappers/javascript/anoncreds-react-native/src/ReactNativeAnoncreds.ts
```diff
@@ -150,12 +172,26 @@ export class ReactNativeAnoncreds implements Anoncreds {
...
+    const selfAttestValues = Object.values(options.selfAttest)
     const schemaKeys = Object.keys(options.schemas)
     const schemaValues = Object.values(options.schemas).map((o) => o.handle)
     const credentialDefinitionKeys = Object.keys(options.credentialDefinitions)
     const credentialDefinitionValues = Object.values(options.credentialDefinitions).map((o) => o.handle)
+
+    const credentialEntries = options.credentials.map((value) => ({
+      credential: value.credential.handle,
+      timestamp: value.timestamp ?? -1,
+      rev_state: value.revocationState?.handle ?? 0,
```
TimoGlastra, line 185: We're using camelCase everywhere, but not for `rev_state`?

TimoGlastra (approved):

wrappers/javascript/anoncreds-react-native/cpp/turboModuleUtility.cpp
genaris: I think it can remain being named `revocationState` because what anoncreds binary receives in the end is the FfiCredentialEntry structure.

wrappers/javascript/anoncreds-react-native/cpp/anoncreds.cpp
genaris: Right. This was based on FFI interface in NodeJs, where it was still present. So I think we'll need to remove it from there as well.

wrappers/javascript/anoncreds-react-native/cpp/anoncreds.cpp
```diff
@@ -129,10 +349,10 @@ jsi::Value createPresentation(jsi::Runtime &rt, jsi::Object options) {
   auto credentialsProve =
       jsiToValue<FfiList_FfiCredentialProve>(rt, options, "credentialsProve");
   auto selfAttestedNames =
-      jsiToValue<FfiStrList>(rt, options, "selfAttestedNames");
+      jsiToValue<FfiStrList>(rt, options, "selfAttestNames");
```
blu3beri, line 352: I think we should change it back to `selfAttestedNames` and make the API take those keys.

wrappers/javascript/anoncreds-react-native/cpp/turboModuleUtility.cpp
```diff
@@ -255,10 +258,13 @@ jsiToValue<FfiList_FfiCredentialProve>(jsi::Runtime &rt, jsi::Object &options,
       auto isPredicate = jsiToValue<int8_t>(rt, valueAsObject, "isPredicate");
       auto reveal = jsiToValue<int8_t>(rt, valueAsObject, "reveal");
 
+      char *ffiStr = new char[sizeof(referent)];
```
blu3beri, line 261: Any specific reason for this change?

blu3beri (approved):

wrappers/javascript/anoncreds-react-native/cpp/turboModuleUtility.cpp
genaris: Yes, it was not working if passed directly to the struct. Probably also related with that automatic wiping that is happening.

wrappers/javascript/anoncreds-react-native/cpp/anoncreds.cpp
genaris: I see this method is still exposed in FFI, so I think we can keep it even if not used in regular API flow.

wrappers/javascript/anoncreds-react-native/cpp/anoncreds.cpp
genaris: Not sure what you mean by this. I only changed the wording from "attested" to "attest" to match other classes such as the one in NodeJs and the FFI method.

blu3beri: I mean that `selfAttestedNames` is grammatically correct and `selfAttestNames` is not. But if the other is used more than we should use that one for consistency.

Issue #125, opened 2023-03-02
Add `nonRevokedIntervalOverrideList` to the React Native wrapper


blu3beri:
This is required for `verifyPresentation`

PR #131, opened 2023-03-06
fix(js): several fixes for js wrapper

TimoGlastra:
Similar to https://github.com/hyperledger/indy-vdr/pull/170 and https://github.com/hyperledger/aries-askar/pull/111 but for anoncreds

### DIFF
--- a/wrappers/javascript/anoncreds-shared/src/Anoncreds.ts
+++ b/wrappers/javascript/anoncreds-shared/src/Anoncreds.ts
@@ -52,7 +52,11 @@ export interface Anoncreds {
     issuerId: string
     signatureType: string
     supportRevocation: boolean
-  }): { credentialDefinition: ObjectHandle; credentialDefinitionPrivate: ObjectHandle; keyProof: ObjectHandle }
+  }): {
+    credentialDefinition: ObjectHandle
+    credentialDefinitionPrivate: ObjectHandle
+    keyCorrectnessProof: ObjectHandle
+  }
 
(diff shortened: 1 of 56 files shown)


wrappers/javascript/anoncreds-shared/src/Anoncreds.ts
```diff
@@ -39,6 +39,7 @@ export interface Anoncreds {
   version(): string
 
   getCurrentError(): string
+  setDefaultLogger(): void
```
genaris, line 42: Is there a way in these new components to set a custom logger like we did in indy-sdk to log in our language of choice?

TimoGlastra: We never implemented that method for React Native, only Node.JS. I think it should be possible, but not sure how to do it currently. We thought defaultLogger would be enough for now and we can add the custom logger later on

genaris: I didn't do that recently but I remember that in NodeJS I needed to set environment variable `RUST_LOG=anoncreds:loglevel` in addition to a call to set_default_logger.  In React Native this can also be achieved so Rust traces can be seen in Android logcat or something?

blu3beri:
closes #125

TimoGlastra:
Tested in React Native

blu3beri:
Since we can just use `getBool`, this closes #141

blu3beri (approved):

PR #137, opened 2023-03-09
fix: add support for legacy schema and cred def id

blu3beri:
@genaris this should fix your issue regarding the invalid cred_def / schema identifiers.

I was wondering if anyone knows a regex to match the revocation registry definition id and the revocation registry id.

Since the new identifier checker checks if it's just a uri, we will validate incorrect schemas with a version of `3` for example. Curious to hear if we want to fix this or leave as-is.

### DIFF
--- a/src/data_types/macros.rs
+++ b/src/data_types/macros.rs
@@ -2,7 +2,10 @@
 macro_rules! impl_anoncreds_object_identifier {
     ($i:ident) => {
         use $crate::error::ValidationError;
-        use $crate::utils::validation::{Validatable, LEGACY_IDENTIFIER, URI_IDENTIFIER};
+        use $crate::utils::validation::{
+            Validatable, LEGACY_CRED_DEF_IDENTIFIER, LEGACY_DID_IDENTIFIER,
+            LEGACY_SCHEMA_IDENTIFIER, URI_IDENTIFIER,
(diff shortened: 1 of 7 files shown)


src/data_types/macros.rs
```diff
@@ -29,17 +40,29 @@ macro_rules! impl_anoncreds_object_identifier {
 
         impl Validatable for $i {
             fn validate(&self) -> Result<(), ValidationError> {
+                let legacy_regex = match stringify!($i) {
+                    "IssuerId" => &LEGACY_DID_IDENTIFIER,
+                    "CredentialDefinitionId" => &LEGACY_CRED_DEF_IDENTIFIER,
+                    "SchemaId" => &LEGACY_SCHEMA_IDENTIFIER,
+                    // TODO: we do not have correct validation for a revocation registry and definition id
+                    "RevocationRegistryId" => &LEGACY_DID_IDENTIFIER,
```
TimoGlastra, line 48: Both are the same (rev reg id and rev reg def id). You can find regexes for this in identifiers.ts in afj

TimoGlastra (approved):

TimoGlastra:
Already going to merge this, but would be good if you can make another pr with the changes for rev regs

PR #153, opened 2023-03-20
fix: rename master secret to link secret

blu3beri:


src/data_types/cred_request.rs
```diff
@@ -11,8 +11,10 @@ pub struct CredentialRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     prover_did: Option<String>,
     cred_def_id: CredentialDefinitionId,
-    pub blinded_ms: ursa::cl::BlindedCredentialSecrets,
-    pub blinded_ms_correctness_proof: ursa::cl::BlindedCredentialSecretsCorrectnessProof,
+    #[serde(alias = "blinded_ms")]
+    pub blinded_ls: ursa::cl::BlindedCredentialSecrets,
+    #[serde(alias = "blinded_ms_correctness_proof")]
+    pub blinded_ls_correctness_proof: ursa::cl::BlindedCredentialSecretsCorrectnessProof,
```
blu3beri, line 17: This adds compatibility for both `blinded_ms` and `blinded_ls` if the spec changes this field.

src/data_types/link_secret.rs
```diff
@@ -0,0 +1,56 @@
+use std::fmt;
+
+use crate::error::ConversionError;
+use serde::{Deserialize, Serialize};
+use ursa::cl::{prover::Prover as UrsaProver, MasterSecret};
+
+#[derive(Serialize, Deserialize)]
+pub struct LinkSecret {
+    pub value: MasterSecret,
```
blu3beri, line 9: The structure is still:

```json
{
  "value": { ms: "BIGNUMBER" }
}
```

genaris: Is it needed to keep this structure? I didn't see it referenced in the spec and I think it would be a good opportunity to simplify it to avoid confusions (IMO it should be simply a plain string)

blu3beri: It's internal so I don't think it matters. Changing this should be doable with serde remote, but will be done if it's required.

blu3beri: > I think it does matter, because when dealing with the library we usually call `load` or `toJson`, which in this case is misleading because we don't have any reference regarding how the expected format should be. In mi opinion these methods do make sense for standardized fields (i.e. the ones whose structure is clearly stated in AnonCreds spec).
> 
> As a direct user of this library, I can think of treating it as an opaque string and don't care about it's internal structure. However, if I want to use a Link Secret generated by other means or if I want to use one generated here for another purpose I wouldn't expect it to be a JSON object with such particular structure.
> 
> It feels more natural and logic to me if this `load` that receives a string would simply receive the Link Secret value instead of a JSON structure. Likewise, the `toJson` that outputs a string might become `toString`.

I agree with the structure feeling incorrect from a to/from json perspective. But `fromJson` taking a string as input feels very inconsistent with the API and every other implementation. 

I think reworking the structure to this:

```json
{
   "link_secret": "BIGNUMBER"
}
```

would be good. I think avoiding the `link_secret` in the JSON, by using a `fromString` method, would only cause confusing with regards to the rest of the API.

src/data_types/cred_request.rs
```diff
@@ -298,15 +300,15 @@ mod cred_req_tests {
...
         let credential_offer = credential_offer(correctness_proof, true)?;
 
         let res = create_credential_request(
             None,
             PROVER_DID,
             &cred_def,
-            &master_secret,
-            MASTER_SERCET_ID,
+            &link_secret,
+            LINK_SERCET_ID,
```
genaris, line 311:
```suggestion
            LINK_SECRET_ID,
```

src/data_types/cred_request.rs
blu3beri: This does however create the output of `blinded_ls` and not `blinded_ms` anymore.

src/data_types/link_secret.rs
blu3beri: Latest changes now include dealing with the link secret as a sting in decimal layout (like in the datamodels).

src/data_types/cred_request.rs
TimoGlastra: Didn't think about the credential request. Not sure if we should change it now, as it's two more fields to update :/ and that we need to support both ways for backwards compatibility

blu3beri: With alias we support both for incoming data models, but we always output with `ls`. 

I mainly changed it because blinded_ms that refers to a link secret feels weird. 

Do you think the implications outweigh the naming convention?

blu3beri: Oh that's a difficult one, I would really like it to be `ls`. Changing it later would be a breaking change and it feels like an oversight. If we could do it later without a breaking change, I am 100% to just leave it for now.

genaris: Anoncreds v1 is expected to keep compatibility with what has been done, so it's not likely to change this field. At least not in a breaking manner.

TimoGlastra:
@blu3beri is this ready-ish to merge or is there any outstanding disucssions here?

blu3beri:
> @blu3beri is this ready-ish to merge or is there any outstanding disucssions here?

Code is good, but `yarn` is acting very weird in the javascript tests, not sure what is going on there.

src/data_types/cred_request.rs
genaris: @blu3beri This is not solved. Unless `Sercet` means something...

wrappers/python/anoncreds/__init__.py
```diff
@@ -41,7 +41,6 @@
     "CredentialRequest",
     "CredentialRequestMetadata",
     "MasterSecret",
-    "NonrevokedIntervalOverride",
```
TimoGlastra, line 44: Why is this removed?

wrappers/python/demo/test.py
```diff
@@ -135,113 +128,14 @@
...
-
-present.add_attributes(
-    recv_cred,
-    "attr2_referent",
-    reveal=False,
-    timestamp=time_revoke_cred,
-    rev_state=rev_state
-)
-
-present.add_attributes(
```
TimoGlastra, line 238: Why removed?

TimoGlastra (approved):
See comment from ariel also

wrappers/python/demo/test.py
blu3beri: Ah I think an issue occurred with mergin. Will revert it.

PR #160, opened 2023-03-22
Remove go wrapper from the library

Vickysomtee:


blu3beri (approved):

TimoGlastra:
We should set-up some requirements for ci jobs to pass, that way we can enable auto-merge!

PR #166, opened 2023-03-30
chore(python): added pyproject.toml

### LINKED ISSUE #154: Need pyproject.toml to publish
#117 is blocking #152. [The wrappers/python needs a pyproject.toml](https://github.com/hyperledger/aries-acapy-tools/blob/main/pyproject.toml) (like the one linked) so that I can generate the token, etc etc.

blu3beri:
closes #154 

@dbluhm any chance you can review this? I followed some guides, and the demo is working, but python setup still remains a small mystery to me :).

### DIFF
--- a/wrappers/python/pyproject.toml
+++ b/wrappers/python/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools>=61.0.0", "wheel"]
+build-backend = "setuptools.build_meta"


wrappers/python/demo/test.py
```diff
@@ -25,250 +25,251 @@
...
-        "sex": "male",
-        "name": "Alex",
-        "height": "175",
-        "age": "28"
-    },
-    None,
-    rev_reg_id,
-    revocation_status_list,
-    CredentialRevocationConfig(
+def run():
```
blu3beri, line 28: Nothing changed in this file except it being in a `run` function. Which I did for it to be able to be run with `poetry run demo` which I thought was nice. Maybe there is a nicer way or a more idiomatic way. Maybe with `__main__`?

TimoGlastra (approved):
LGTM, but as you said would be good if @dbluhm can also take a look at it

dbluhm:
I'm not super familiar with how bundling binaries looks in pyproject.toml so I'll read up a bit and review! (Looks good to me though)

dbluhm:
Did some digging; it would seem that poetry is not currently well suited to building platform specific wheels. It might be possible but the `poetry build` command can only build pure python wheels right now. @ryjones could you provide a little more detail on why a pyproject.toml specifically is required to get PyPI tokens? We can set up the pyproject.toml to configure setuptools (which looks like the best option for building platform specific wheels) instead of poetry but I'm not sure that will fit the workflow you were hoping for.

Input from @andrewwhitehead (who has far more experience building python packages with bundled binaries) would probably also be helpful, if you get a chance.

ryjones:
@dbluhm I was following this guide: https://realpython.com/pypi-publish-python-package/#configure-your-package

dbluhm:
@blu3beri I propose we construct a minimal pyproject.toml with just:

```
# pyproject.toml

[build-system]
requires = ["setuptools>=61.0.0", "wheel"]
build-backend = "setuptools.build_meta"
```

And continue using setuptools for builds, following the same pattern as askar and indy-shared-rs.

blu3beri:
Thanks for the work @dbluhm! I will change the PR and revert the setup.py and double check with askar.

PR #172, opened 2023-04-01
feat(js): revocation status list from json

TimoGlastra:
This add support to create a revocation status list from json in the JS wrapper. The method was already availalbe in the ffi layer as `anoncreds_revocation_list_from_json`, but I changed it to `anoncreds_revocation_status_list_from_json` as that aligns more with the rest of the naming. If we choose to call it `revocation_list` we should change it everywhere at once.

Also fixes a few small issues in the RN wrapper.


TimoGlastra:
This is pending a fix to include all `from_json` methods in the header file, as currently the latest anoncreds release for react native is broken due to missing methods

blu3beri (approved):

PR #175, opened 2023-04-01
fix: correctly generate header

blu3beri:


blu3beri:
Hmm, I see now that we do something like this:

```cpp
auto json = jsiToValue<std::string>(rt, options, "json");

  ObjectHandle out;
  ByteBuffer b = stringToByteBuffer(json);

  ErrorCode code = anoncreds_revocation_registry_from_json(b, &out);
  auto returnValue = createReturnValue(rt, code, &out);
```

which should be 

```cpp
auto json = jsiToValue<ByteBuffer>(rt, options, "json");

  ObjectHandle out;

  ErrorCode code = anoncreds_revocation_registry_from_json(json, &out);
  auto returnValue = createReturnValue(rt, code, &out);
```

it does not change a lot of the code, but this is more inline way of doing it. (not an issue for now, but we can clean this up in the future).

TimoGlastra (approved):

PR #186, opened 2023-04-06
fix(js): revocation status list typo and optional timestamp

genaris:
Some things I found while digging a bit with revocation status lists in JavaScript.

Also, maybe more for a discussion or separate issue, but I find a bit odd to have both `updateTimestamp()` and `update()` methods in `RevocationStatusList` class from JS API. It could be more straightforward to just have `update()` and, if the only defined parameter is the timestamp, it can call `anoncreds.updateRevocationStatusListTimestampOnly`.

### DIFF
--- a/wrappers/javascript/anoncreds-react-native/src/ReactNativeAnoncreds.ts
+++ b/wrappers/javascript/anoncreds-react-native/src/ReactNativeAnoncreds.ts
@@ -180,7 +180,6 @@ export class ReactNativeAnoncreds implements Anoncreds {
   }
 
   public createLinkSecret(): string {
-    4
     return handleError(anoncredsReactNative.createLinkSecret({}))
   }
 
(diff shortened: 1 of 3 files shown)


wrappers/javascript/anoncreds-react-native/src/ReactNativeAnoncreds.ts
```diff
@@ -176,6 +180,7 @@ export class ReactNativeAnoncreds implements Anoncreds {
   }
 
   public createLinkSecret(): string {
+    4
```
TimoGlastra, line 183:
```suggestion
```

TimoGlastra (approved):

TimoGlastra:
> Also, maybe more for a discussion or separate issue, but I find a bit odd to have both updateTimestamp() and update() methods in RevocationStatusList class from JS API. It could be more straightforward to just have update() and, if the only defined parameter is the timestamp, it can call anoncreds.updateRevocationStatusListTimestampOnly.

That makes sense to me!

PR #199, opened 2023-05-01
chore: lower ios compatibility

blu3beri:
Now using static libraries, which have a lower minimum iOS version suppport, does increase download and possible bundle size (last one is untested).


TimoGlastra (approved):
I think for now the increased bundle size is fine. Maybe we can allow the user to choose in the future, or just wait for iOS 16 to have 95%+ support

Issue #203, opened 2023-05-10
Ursa EOL


berendsliedrecht:
[Ursa](https://github.com/hyperledger/ursa) has been moved to EOL status. This means that Anoncreds-rs has to move away from this library as it will not be updated or will get any security fixes.

There is some ongoing work for extracting the work, most notably for this library is [anoncreds-clsignatures](https://github.com/hyperledger/anoncreds-clsignatures-rs).

Looking at the codebase, with `rg --pcre2 'ursa(?!::cl).*' .` shows only some dependency `Ursa::bn`, which can be replaced with openssl (native depencency) / BigNum (pure rust). Everything else, which is in the `cl` package from `Ursa` should be moved to the `anoncreds-clsignatures-rs` repository, AFAIK.

If this is the only work, moving away form Ursa would be rather easy as long as anoncreds-clsignatures is finished.

@andrewwhitehead Do see any additional work or is really almost everything covered by `anoncreds-clsignatures`?

TimoGlastra:
Would moving to openssl resolve some of the issues around anoncreds being slow in WASM?

blu3beri:
> Would moving to openssl resolve some of the issues around anoncreds being slow in WASM?

No, that would need to be done in `anoncreds-clsignatures` as it is slow because of generating some 1024 bit safe primes for the credential definition. 

The person behind libsodium has openssl working for WASM, https://github.com/jedisct1/openssl-wasm, but I am not sure about the build process of including this into rusts openssl-src/sys package so we can use it.

And possibly it is not even faster to use openssl-wasm instead of a pure rust as AFAIK the openssl version is quite faster because half of the big number implementation of openssl is done in assembly. Which would then be done with some polyfill / pure webassembly and I am not aware of any benchmarks between them.

andrewwhitehead:
> @andrewwhitehead Do see any additional work or is really almost everything covered by `anoncreds-clsignatures`?

I believe that should be the only (direct) dependency needed. If you're using the big number support directly then we may have to check that it's exposed.

The 'native' backend currently uses glass-pumpkin, but the prime generation is too slow to be practical (last I checked). A recent alternative might be [crypto-primes](https://crates.io/crates/crypto-primes) but I haven't benchmarked that yet. There is also [num-bigint-dig](https://crates.io/crates/num-bigint-dig).

PR #207, opened 2023-05-17
chore: use crypto-bigint instead of ursa

### LINKED ISSUE #202: Zeroing the Link Secret
I raised a question on Discord and it was recommended I raise it as an issue to track here. Original messasge pasted below:

> I was reading through this issue https://github.com/hyperledger/anoncreds-rs/issues/194 and had a question:
>
> With the link secret (formerly master secret) now being returned as just a plain old string value, should we be thinking about zeroing out memory where the secret was held? I'm not in the habit of thinking about this in python but that was perhaps one advantage of having the rust layer hold the value so that it could then zero the memory when it was released.

@blu3beri responded to my question on Discord indicating that zeroing the link secret is probably something we want to do but there are some complexities that arise from the fact that it's a BigNumber type from Ursa which doesn't implement zeroize.

berendsliedrecht:
closes #202 
partially `resolves` #203

### DIFF
--- a/src/data_types/link_secret.rs
+++ b/src/data_types/link_secret.rs
@@ -1,28 +1,18 @@
 use crate::error::ConversionError;
+use crypto_bigint::{self, rand_core::OsRng, Encoding, Random, U256};
+use serde::Serialize;
+use zeroize::Zeroize;
 use std::fmt;
-use ursa::{
-    bn::BigNumber,
-    cl::{prover::Prover as UrsaProver, MasterSecret},
-};
-pub struct LinkSecret(pub ursa::bn::BigNumber);
+use ursa::cl::MasterSecret;
+
+// TODO: This should serialize to decimal and not lower le hex
+#[derive(Zeroize, Clone, Copy, Serialize)]
+pub struct LinkSecret(pub U256);
 
 impl LinkSecret {
-    pub fn new() -> Result<Self, ConversionError> {
-        let value = UrsaProver::new_master_secret()
-            .and_then(|v| v.value())
-            .map_err(|err| {
-                ConversionError::from_msg(format!("Error creating link secret: {err}"))
-            })?;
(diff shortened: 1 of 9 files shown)


src/data_types/link_secret.rs
```diff
@@ -46,36 +36,28 @@ impl TryInto<MasterSecret> for LinkSecret {
...
-impl TryFrom<&str> for LinkSecret {
-    type Error = ConversionError;
-
-    fn try_from(value: &str) -> Result<Self, Self::Error> {
-        Ok(Self(BigNumber::from_dec(value).map_err(|err| {
-            ConversionError::from_msg(format!("Error creating link secret: {err}"))
-        })?))
+impl From<&str> for LinkSecret {
+    fn from(value: &str) -> Self {
+        Self(U256::from_be_hex(value))
```
andrewwhitehead, line 60: This will panic on bad input, which doesn't seem ideal. I've been decoding the hex to bytes and using U256::from_be_slice but crypto-bigint could use fallible methods for this.

blu3beri:
@andrewwhitehead I have been running into some small issues between converting a U256 bigint to an ursa master secret.

Right now, I am doing quite a hacky way of serializing a link_secret, U256, into json and deserializing it into a master_secret, openssl bignum. Openssls bignum in JSON is represented as a decimal, but the U256 defaults to lower_le_hex.

Do you know of a way to fix this? There are some libraries which can convert between hex strings and decimal strings, which might be fine for now as this is a very temporary solution, but I might be missing some better conversion.

andrewwhitehead:
Added a PR for decimal support, although it might still need some changes: https://github.com/RustCrypto/crypto-bigint/pull/246

berendsliedrecht:
@andrewwhitehead #237 makes this redundant, correct?

PR #213, opened 2023-05-24
build(android): use custom cross images

blu3beri:
- Using custom images for building with lower Android support
- If this fully works the same will be done with https://github.com/hyperledger/aries-askar and https://github.com/hyperledger/indy-vdr

### DIFF
--- a/Cross.toml
+++ b/Cross.toml
@@ -5,13 +5,13 @@ image = "ghcr.io/rust-cross/manylinux2014-cross:aarch64"
 image = "ghcr.io/rust-cross/manylinux2014-cross:x86_64"
 
 [target.x86_64-linux-android]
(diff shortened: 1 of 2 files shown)


Cross.toml
```diff
@@ -5,13 +5,13 @@ image = "ghcr.io/rust-cross/manylinux2014-cross:aarch64"
 image = "ghcr.io/rust-cross/manylinux2014-cross:x86_64"
 
 [target.x86_64-linux-android]
-image = "ghcr.io/cross-rs/x86_64-linux-android:main"
+image = "ghcr.io/hyperledger/aries-builder-images/cross/x86_64-linux-android:main"
```
TimoGlastra, line 8: Nice!

TimoGlastra (approved):

PR #218, opened 2023-06-02
Feat/kotlin wrapper

TheTreek:
Created a Kotlin MPP wrapper that should have the same functionality as the existing node.js wrapper.


TheTreek:
Use https://github.com/hyperledger/aries-uniffi-wrappers/pull/11 instead.

Issue #224, opened 2023-07-17
Replace the current ursa dependency with AnonCreds clsignatures


swcurran:
Update the current Ursa dependency in this repository with [AnonCreds CLSignatures](https://github.com/hyperledger/anoncreds-clsignatures-rs).  The API is mostly the same, although there are some differences, particularly in the use of the tails file. 

The API for those calling the AnonCreds RS library should not change at all -- although again, feel free to propose changes where appropriate, such as with the tails file.  For example, where the tails file is no longer needed, perhaps keep it in the API, but mark it as optional/deprecated.

swcurran:
FYI - @amanji , @esune

swcurran:
FYI -- @TimoGlastra @blu3beri @genaris

Issue #221, opened 2023-07-10
Update object finalizers and add keepalives - from indy-shared-rs


swcurran:
The changes from this PR in the Python wrapper of indy-shared-rs needs to be done in this repo — https://github.com/hyperledger/indy-shared-rs/pull/26/files.  A new release should be triggered for the Python wrapper immediately after.  This is to address a bug in the verifiers using anoncreds-rs with the Python wrapper — verification failing when multiple revocation registries are used.

swcurran:
@amanji -- this is a Python task as discussed last Friday.  The Python wrapper for this repo should be about the same (give or take :-) ) as indy-shared-rs, so this should be a straight forward task, even though there is little detail. Do check on a test -- I didn't check today to be sure, but I recall that being one.

PR #226, opened 2023-07-20
chore: replace ursa lib with anoncreds-clsignatures-rs

amanji:
Replaces `ursa` library with `anoncreds-clsignatures-rs`.
Removes some external dependencies which are now packaged with the updated library.
Resolves: https://github.com/hyperledger/anoncreds-rs/issues/221
Resolves: https://github.com/hyperledger/anoncreds-rs/issues/224

Many thanks to @andrewwhitehead for help on this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,9 @@
 [package]
 name = "anoncreds"
 version = "0.1.0"
-authors = ["Hyperledger AnonCreds Contributors <anoncreds@lists.hyperledger.org>"]
+authors = [
+    "Hyperledger AnonCreds Contributors <anoncreds@lists.hyperledger.org>",
+]
 description = "Verifiable credential issuance and presentation for Hyperledger AnonCreds (https://www.hyperledger.org/projects), which provides a foundation for self-sovereign identity."
 edition = "2021"
 license = "Apache-2.0"
@@ -32,13 +34,14 @@ rand = "0.8.5"
 regex = "1.7.1"
 serde = { version = "1.0.155", features = ["derive"] }
 bitvec = { version = "1.0.1", features = ["serde"] }
-serde_json = { version = "1.0.94", features = ["raw_value"]}
+serde_json = { version = "1.0.94", features = ["raw_value"] }
 sha2 = "0.10.6"
 tempfile = "3.4.0"
 thiserror = "1.0.39"
-ursa = { version = "0.3.7", default-features = false, features = ["cl_native", "serde"] }
 zeroize = { version = "1.5.7", optional = true, features = ["zeroize_derive"] }
+anoncreds-clsignatures = "0.1"
 
+# TODO: Can we remove this since ursa is no longer a dependency?
 # We add the openssl dependency here because ursa does not expose a vendored openssl feature
 # Since we use "cl_native" as a feature, which uses openssl, we can add a vendored build with 
 # the new exposed "vendored" feature
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ crate-type = ["staticlib", "rlib", "cdylib"]
 default = ["ffi", "logger", "zeroize"]
 ffi = ["ffi-support"]
 logger = ["env_logger"]
+vendored = ["anoncreds-clsignatures/openssl_vendored"]
 
 [dependencies]
 bs58 = "0.4.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,6 @@ crate-type = ["staticlib", "rlib", "cdylib"]
 default = ["ffi", "logger", "zeroize"]
 ffi = ["ffi-support"]
 logger = ["env_logger"]
-vendored = ["openssl", "openssl/vendored"]
 
 [dependencies]
 bs58 = "0.4.0"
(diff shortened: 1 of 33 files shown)


src/services/issuer.rs
```diff
@@ -544,12 +542,14 @@ pub fn update_revocation_status_list(
     let tails_reader = TailsFileReader::new_tails_reader(&rev_reg_def.value.tails_location);
     let max_cred_num = rev_reg_def.value.max_cred_num;
 
-    CryptoIssuer::update_revocation_registry(
+    // TODO: need to pass in credential_pub_key and rev_key_priv
+    Issuer::update_revocation_registry(
```
amanji, line 546: @andrewwhitehead

src/services/issuer.rs
```diff
@@ -366,12 +363,13 @@ pub fn create_revocation_status_list(
         let tails_reader = TailsFileReader::new_tails_reader(&rev_reg_def.value.tails_location);
         let issued = (1..=max_cred_num).collect::<BTreeSet<_>>();
 
-        CryptoIssuer::update_revocation_registry(
+        Issuer::update_revocation_registry(
```
amanji, line 366: @andrewwhitehead

src/services/issuer.rs
```diff
@@ -155,8 +152,8 @@ where
         issuer_id,
         tag: tag.to_owned(),
         value: CredentialDefinitionData {
-            primary: credential_public_key.get_primary_key()?.try_clone()?,
-            revocation: credential_public_key.get_revocation_key()?,
+            primary: credential_public_key.get_primary_key().try_clone()?,
+            revocation: credential_public_key.get_revocation_key().cloned(),
```
amanji, line 156: @andrewwhitehead This is optional in the type definition. Is it sufficient to just use `clone` like this?

src/error.rs
```diff
@@ -147,20 +147,20 @@ impl From<serde_json::Error> for Error {
...
-            UrsaCryptoErrorKind::InvalidStructure | UrsaCryptoErrorKind::InvalidParam(_) => {
-                ErrorKind::Input
-            }
-            UrsaCryptoErrorKind::IOError => ErrorKind::IOError,
-            UrsaCryptoErrorKind::InvalidRevocationAccumulatorIndex => ErrorKind::InvalidUserRevocId,
-            UrsaCryptoErrorKind::RevocationAccumulatorIsFull => ErrorKind::RevocationRegistryFull,
-            UrsaCryptoErrorKind::ProofRejected => ErrorKind::ProofRejected,
-            UrsaCryptoErrorKind::CredentialRevoked => ErrorKind::CredentialRevoked,
+            CryptoErrorKind::InvalidState => ErrorKind::InvalidState,
+            // CryptoErrorKind::InvalidStructure | CryptoErrorKind::InvalidParam(_) => {
```
amanji, line 156: @andrewwhitehead I've currently commented these out since they aren't defined in `anoncreds-clsignatures-rs`. I can issue a PR to add those `ErrorKind`s there if we want

Cargo.toml
```diff
@@ -32,13 +34,14 @@ rand = "0.8.5"
...
-serde_json = { version = "1.0.94", features = ["raw_value"]}
+serde_json = { version = "1.0.94", features = ["raw_value"] }
 sha2 = "0.10.6"
 tempfile = "3.4.0"
 thiserror = "1.0.39"
-ursa = { version = "0.3.7", default-features = false, features = ["cl_native", "serde"] }
 zeroize = { version = "1.5.7", optional = true, features = ["zeroize_derive"] }
+anoncreds-clsignatures = "0.1"
 
+# TODO: Can we remove this since ursa is no longer a dependency?
```
andrewwhitehead, line 44: `anoncreds-clsignatures` has an `openssl_vendored` feature that you can use instead.

src/error.rs
andrewwhitehead: They aren't raised in `anoncreds-clsignatures`, so no need to worry about them :)

amanji: Ok will remove them then

src/services/issuer.rs
andrewwhitehead: `cloned` is a method of `Option<&T>`, it converts to `Option<T>` (where T: Clone).

src/services/issuer.rs
andrewwhitehead: I think this method will want `cred_def: &CredentialDefinition` and `rev_reg_priv: &RevocationRegistryPrivate` arguments. It's not necessary to create a tails reader. This method behaves similarly: https://github.com/hyperledger/indy-shared-rs/blob/8f51b31039dd0eff1443d3ed3872addcf0958e4c/indy-credx/src/services/issuer.rs#L247

Cargo.toml
```diff
@@ -20,7 +22,6 @@ crate-type = ["staticlib", "rlib", "cdylib"]
 default = ["ffi", "logger", "zeroize"]
 ffi = ["ffi-support"]
 logger = ["env_logger"]
-vendored = ["openssl", "openssl/vendored"]
```
andrewwhitehead, line 23: I believe this flag can be preserved as `vendored = ["anoncreds-clsignatures/openssl_vendored"]`

swcurran:
Note about this on the interface changes made to those using this library:

- The call to issue a credential no longer requires the tails file as a parameter.
- The call to revoke a credential no longer requires the tails file as a parameter, adds the CredDefinition and the revocation registry private key as parameters.

genaris (commented):
I think that the JS part is fine. It misses the updates for React Native wrapper though. Not a common use case of course (and not sure if current issuance/revocation methods are properly working as we didn't test them systematically), so won't consider it as a blocker if not added to this PR.

amanji:
Yes I should have mentioned that the RN wrappers were not updated. I assume they could be updated in a follow up PR to this.

andrewwhitehead:
anoncreds-clsignatures 0.2 is out now with a few more updates, but I can add another PR for that. I think the revocation status list creation could be optimized so I'll look at that at the same time.

swcurran:
> anoncreds-clsignatures 0.2 is out now with a few more updates, but I can add another PR for that. I think the revocation status list creation could be optimized so I'll look at that at the same time.

@andrewwhitehead — what is the action here?  Can we merge this, do we need to update to 0.2?  Your note implies that we can do the 0.2 update later, but then you didn’t approve this.  Thanks!

andrewwhitehead (approved):

PR #231, opened 2023-08-12
fix: JS wrapper updates

genaris:
Some updates to JavaScript wrappers to match main `branch` status:

- update React Native wrapper to match changes merged after #226. It also fixes an error in `UpdateRevocationStatusList` for NodeJS
- update both NodeJs and React Native according to #229. In that PR I've seen that `FfiCredRevInfo` structure has not been changed, even if `tails_path` is not used anymore. @andrewwhitehead is this on purpose or should be removed from there as well?

I went a bit more further and updated minimum node version to 18, which is in fact the minimum _usable_ version due to the known performance issue of `ref-napi` (which is patched). node 16 will reach its EOL next month anyway.


wrappers/javascript/package.json
```diff
@@ -55,7 +55,10 @@
     "tsconfig-paths": "4.1.0",
     "typescript": "4.5.5"
   },
+  "resolutions": {
+    "ref-napi": "npm:@2060.io/ref-napi"
```
TimoGlastra, line 59: Can't we add this as a direct dependency for all the wrappers and not require the resolution anymore?

genaris: I'm not sure if that would work because it is also an indirect dependency: `ffi-napi` requires it. In such case maybe we can also patch it to depend on the patched `ref-napi` as well.

genaris: @TimoGlastra what about these latest changes? I needed some tweaks for types, but at the end seems to work fine. I think it's worth to put some effort on this considering the several problems users had in setting up these components for NodeJS.

wrappers/javascript/anoncreds-nodejs/test/bindings.test.ts
```diff
@@ -297,7 +297,6 @@ describe('bindings', () => {
         revocationRegistryDefinition,
         revocationRegistryDefinitionPrivate,
         registryIndex: 9,
-        tailsPath,
```
TimoGlastra, line 300: tailsPath is not used anymore (eslint warning)

TimoGlastra (approved):
This is great! Would definitely help in making it easier to use these libraries (we should probably do it for all libraries).

Maybe we can also look at not hosting it under 2060 npm scope, but that's something we can look at later on (not sure what would be a good home)

PR #233, opened 2023-08-16
Revocation registry cleanups

andrewwhitehead:
Updates to anoncreds-clsignatures 0.2.1 and simplifies revocation registry handling.

### DIFF
--- a/src/data_types/rev_status_list.rs
+++ b/src/data_types/rev_status_list.rs
@@ -19,8 +19,8 @@ pub struct RevocationStatusList {
     revocation_list: bitvec::vec::BitVec,
     #[serde(
         rename = "currentAccumulator",
-        skip_serializing_if = "Option::is_none",
-        with = "serde_opt_accumulator"
+        alias = "accum",
+        skip_serializing_if = "Option::is_none"
     )]
     accum: Option<Accumulator>,
     #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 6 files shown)


berendsliedrecht (approved):

src/data_types/rev_status_list.rs
```diff
@@ -17,35 +17,27 @@ pub struct RevocationStatusList {
     issuer_id: IssuerId,
     #[serde(with = "serde_revocation_list")]
     revocation_list: bitvec::vec::BitVec,
-    #[serde(rename = "currentAccumulator", skip_serializing_if = "Option::is_none")]
-    registry: Option<CLSignaturesRevocationRegistry>,
+    #[serde(
+        rename = "currentAccumulator",
+        skip_serializing_if = "Option::is_none",
+        with = "serde_opt_accumulator"
```
bobozaur, line 23: This seems like an overkill. Serialization doesn't seem to require anything custom, so the attribute could instead be `deserialize_with`. But then the custom deserialization seems to only want to achieve having multiple names for the field, which can also be done with `#[serde(alias = "currentAccumulator"]`.

Depending on which should be the serialization name for the field, you might want to use `#[serde(rename = "currentAccumulator", alias = "accum"]` instead.

Deserialization of the `Accumulator` also seems to be over complicated, but that's related to `anoncreds-clsignatures`. I'll try to drop a PR there for this.

berendsliedrecht: Yes, makes sense. `#serde(rename = "currentAccumulator")` should suffice here. 

The test `json_rev_list_can_be_deserialized` might also need to be expanded to make sure that the test vector, which includes `currentAccumulator` checks that `des.accum.is_some()`. as `skip_serializing_if="Option::is_none"` is set, it will silently fail if it can not interpret `currentAccumulator` correctly.

andrewwhitehead: I wasn't trying to change the current deserialization, but it's pretty strange. It accepts values like:
```json
{ ...
  "currentAccumulator": {
    "currentAccumulator": "1 000..."
  }
  ...
}
```

bobozaur: Probably because the visitor implements the `visit_map` method.

andrewwhitehead: So do we think that's a mistake and it's just meant to support both currentAccumulator and accum as the field name? In that case it wouldn't need any custom deserialization.

berendsliedrecht: I am fairly sure that it is a mistake. It was a custom implementation because we could not change it as it was a private field within Ursa (it has been a while so there might be more to it).

bobozaur: Well... The changes are introduced in this PR, so I was honestly expecting you to know the answer to that 😅 . To me it just seemed weird so I pointed it out.

andrewwhitehead: I just moved the deserialization code around and changed it to an Accumulator value, the old implementation had the visit_map method.

andrewwhitehead: Updated to just deserialize from a string, which is the format the tests use. I'm a little surprised the bitstring serialization is a list instead of hex bytes or something.

bobozaur (approved):

PR #235, opened 2023-08-18
Remove tails path from FfiCredRevInfo

andrewwhitehead:


wrappers/javascript/anoncreds-shared/src/api/RevocationRegistry.ts
```diff
@@ -4,12 +4,6 @@ import type { RevocationRegistryDefinition } from './RevocationRegistryDefinitio
 import { AnoncredsObject } from '../AnoncredsObject'
 import { anoncreds } from '../register'
 
-export type RevokeCredentialOptions = {
```
berendsliedrecht, line 7: Shouldn't the `tailPath` field just be removed?

andrewwhitehead: I checked and this structure isn't used anywhere

berendsliedrecht: Okay, then it is all good.

berendsliedrecht (approved):

PR #237, opened 2023-08-18
Simplify LinkSecret handling

andrewwhitehead:
This updates to anoncreds-clsignatures 0.2.2 to take advantage of simple conversion from the link secret to a BigNumber. I don't think there's any reason to support conversion from ClLinkSecret to LinkSecret anymore as the `build_credential_values` method is updated to use LinkSecret.

### DIFF
--- a/src/data_types/link_secret.rs
+++ b/src/data_types/link_secret.rs
@@ -1,16 +1,15 @@
-use crate::cl::{bn::BigNumber, LinkSecret as ClLinkSecret, Prover as CryptoProver};
-use crate::error::ConversionError;
 use std::fmt;
 
-pub struct LinkSecret(pub BigNumber);
+use crate::cl::{bn::BigNumber, LinkSecret as ClLinkSecret, Prover as CryptoProver};
+use crate::error::ConversionError;
+
+pub struct LinkSecret(pub(crate) BigNumber);
 
 impl LinkSecret {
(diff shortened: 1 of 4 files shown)


src/data_types/link_secret.rs
```diff
@@ -1,16 +1,15 @@
-use crate::cl::{bn::BigNumber, LinkSecret as ClLinkSecret, Prover as CryptoProver};
-use crate::error::ConversionError;
 use std::fmt;
 
-pub struct LinkSecret(pub BigNumber);
+use crate::cl::{bn::BigNumber, Prover as CryptoProver};
+use crate::error::ConversionError;
+
+pub struct LinkSecret(pub(crate) BigNumber);
```
berendsliedrecht, line 6: Would it be possible to derive `zeroize` on this? It was not done before due to Ursa but it might be possible now.

berendsliedrecht (approved):

src/data_types/link_secret.rs
andrewwhitehead: For version 0.2.2 the crate is using openssl's secure context for big numbers, which as I understand it zeroizes the entire buffer before it's released. So this type could implement ZeroizeOnDrop, but probably not Zeroize.

berendsliedrecht:
resolves #207

Issue #236, opened 2023-08-18
Strings not freed in the JS wrapper?


andrewwhitehead:
I don't see any references to `anoncreds_string_free`, but this is likely required for all strings coming from the FFI layer. This could be a significant memory leak over time.

berendsliedrecht:
Yeah that does not seem to be too good. I think we have three solutions but they all have some disadvantages

1. Expose `string_free` publicly and allow the consumer to deal with it.
  - Downside is that this is quite the uncommon pattern in JS and likely they will still forget some allocations.
2. Within the wrapper copy the value and immediately call it (it might even already be copied `ffi-napi` and c++ side).
  - Would still require all manual frees but at least we control it and not the consumer.
 3. Wait for Typescript 5.2 to be stable as it introduces the `using` keyword which would finally allow a cleanup function to be ran when it goes out-of-scope.
   - The best option by far, but I am unsure of the timeline (https://github.com/microsoft/TypeScript/issues/54298).

IMO it is okay to wait for typeScript 5.2 and leave this as-is for now.

genaris:
> 3\. Wait for Typescript 5.2 to be stable as it introduces the `using` keyword which would finally allow a cleanup function to be ran when it goes out-of-scope.
> 
> * The best option by far, but I am unsure of the timeline ([TypeScript 5.2 Iteration Plan microsoft/TypeScript#54298](https://github.com/microsoft/TypeScript/issues/54298)).

Well, if that iteration plan is accurate, it will be out in 4 days from now! So definitely it's worth waiting a bit and doing not only this but also for object handle clear() methods.

If not, I'd go for option 2.

andrewwhitehead:
I think the wrapper copies these FFI string buffers to Strings anyway, so it should be possible to free the reference before returning. I noticed later that this would apply to `anoncreds_buffer_free` as well, for binary buffers.

PR #238, opened 2023-08-31
fix(js): free native strings

genaris:
Not completely sure if it's the right approach (especially in react-native) but here is a simple attempt to free strings automatically in JS wrappers before returning to the caller, as discussed in https://github.com/hyperledger/anoncreds-rs/issues/236.

### DIFF
--- a/wrappers/javascript/anoncreds-nodejs/src/NodeJSAnoncreds.ts
+++ b/wrappers/javascript/anoncreds-nodejs/src/NodeJSAnoncreds.ts
@@ -724,11 +724,13 @@ export class NodeJSAnoncreds implements Anoncreds {
     this.handleError()
 
     const returnValue = handleReturnPointer<{ data: Buffer; len: number }>(ret)
-    const output = new Uint8Array(byteBufferToBuffer(returnValue))
+    const jsonAsArray = new Uint8Array(byteBufferToBuffer(returnValue))
+
+    const output = new TextDecoder().decode(jsonAsArray)
 
     this.nativeAnoncreds.anoncreds_buffer_free(returnValue.data)
 
-    return new TextDecoder().decode(output)
+    return output
   }
 
   public getTypeName(options: { objectHandle: ObjectHandle }) {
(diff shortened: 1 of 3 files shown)


wrappers/javascript/anoncreds-nodejs/src/NodeJSAnoncreds.ts
```diff
@@ -723,6 +726,8 @@ export class NodeJSAnoncreds implements Anoncreds {
     const returnValue = handleReturnPointer<{ data: Buffer; len: number }>(ret)
     const output = new Uint8Array(byteBufferToBuffer(returnValue))
 
+    this.nativeAnoncreds.anoncreds_buffer_free(returnValue.data)
```
TimoGlastra, line 729: Is a buffer also copied and not a ref?

TimoGlastra (approved):
This doesn't include the RN implementation yet right?

berendsliedrecht (approved):
LGTM. TypeScript `5.2.2` has been released so we can use the new `using` declaration to probably make it even better. This could even be done in the `shared` part of the library fixing it for Node.JS and React Native

genaris:
> This doesn't include the RN implementation yet right?

I couldn't test it deeply but I've updated those template `createReturnValue` functions related to strings (char *) and buffers (ByteBuffer) to call these clean up functions, so the fix should work also for RN.

genaris:
> LGTM. TypeScript `5.2.2` has been released so we can use the new `using` declaration to probably make it even better. This could even be done in the `shared` part of the library fixing it for Node.JS and React Native

Absolutely! I was looking at the release notes and `using` seems really promising for our exact use case (here and in `aries-askar` wrapper as well). But having it available in TS does automatically mean that we can use it in Node.js and React Native? I've seen that the [related proposal](https://github.com/tc39/proposal-explicit-resource-management) is at stage 3 (candidate). Does that mean that it will take some time until JS engines like V8 or Hermes adopt it officially?

wrappers/javascript/anoncreds-nodejs/src/NodeJSAnoncreds.ts
berendsliedrecht: It would depend on how `ffi-napi` `ref-napi` handles this. It seems to me that they copy the data, but I haven't checked the code, yet.

andrewwhitehead (approved):

PR #239, opened 2023-09-05
Simplify create_credential parameters and remove RevocationRegistryId

andrewwhitehead:
As far as I understand, there's no such thing as a RevocationRegistryId, since RevocationRegistryDefinitionId covers that usage (for example, see within the Credential structure). Looking at the unit tests they seem to have been used interchangeably. If anything, I think there would need to be a RevocationStatusListId instead. So this PR removes RevocationRegistryId and just uses RevocationRegistryDefinitionId in its place.

When issuing via `create_credential`, the revocation registry definition ID can be taken from the status list object.

The Python wrapper is updated, but I haven't looked at updating the JS wrappers yet.

### DIFF
--- a/include/libanoncreds.h
+++ b/include/libanoncreds.h
@@ -176,6 +176,7 @@ typedef struct FfiList_FfiStr FfiStrList;
 typedef struct FfiCredRevInfo {
   ObjectHandle reg_def;
   ObjectHandle reg_def_private;
+  ObjectHandle status_list;
   int64_t reg_idx;
 } FfiCredRevInfo;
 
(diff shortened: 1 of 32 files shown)


TimoGlastra (approved):
Changes to the wrapper seem minimal (besides formatting). @genaris do you have time to update this in the JS wrapper?

genaris:
> Changes to the wrapper seem minimal (besides formatting). @genaris do you have time to update this in the JS wrapper?

I always have time for my beloved anoncreds-rs JS wrapper 😄.

genaris:
@andrewwhitehead FYI I submitted a PR to this branch updating JS wrappers according to these new changes: https://github.com/andrewwhitehead/anoncreds-rs/pull/1

PR #249, opened 2023-09-24
feat: switched to pnpm and added react native example app

berendsliedrecht:
PR got a bit bigger than expected.

- Move from yarn to pnpm as it makes dealing with workspaces a lot easier
- Added a React native example app. Right now it does not include an easy way to use locally-built binaries. This should be added in the future, but it can get quite complex.
- Improved the registration of the React Native module a bit, might be worthwhile in the future to see if we can move this to an expo module as it would decrease maintenance a lot.


wrappers/javascript/lerna.json
```diff
@@ -1,7 +1,7 @@
 {
   "version": "0.2.0-dev.1",
   "useWorkspaces": true,
```
TimoGlastra, line 3: I think useWorkspaces was deprecated in lerna 7 right?

wrappers/javascript/packages/anoncreds-react-native/src/register.ts
```diff
@@ -0,0 +1,75 @@
...
+      throw new Error(
+        `Failed to create a new ${libraryName} instance: React Native is not running on-device. ${libraryName} can only be used when synchronous method invocations (JSI) are possible. If you are using a remote debugger (e.g. Chrome), switch to an on-device debugger (e.g. Flipper) instead.`
+      )
+    }
+
+    // Call the synchronous blocking install() function
+    const result = anoncredsModule.install()
+    if (!result)
+      throw new Error(
+        'Failed to create a new MMKV instance: The native MMKV Module could not be installed! Looks like something went wrong when installing JSI bindings.'
```
TimoGlastra, line 64:
```suggestion
        'Failed to create a new Anoncreds instance: The native Anoncreds Module could not be installed! Looks like something went wrong when installing JSI bindings.'
```

TimoGlastra (approved):
I personally like trailing commas, but I'm fine with these changes 👍

PR #253, opened 2023-10-19
fix: do not free string

TimoGlastra:
Based on discussion in this PR" https://github.com/hyperledger/aries-framework-javascript/pull/1606


TimoGlastra:
One thing I'm curious about: Do we also need to modify the calls to string_free in the RN wrapper? I think it's probably different as there's no automatic GC?

genaris:
> One thing I'm curious about: Do we also need to modify the calls to string_free in the RN wrapper? I think it's probably different as there's no automatic GC?

For React Native I think it is good to keep freeing it manually, considering that the only thing we do there is assigning a pointer. In Node.js I'm still not too convinced the way it works, but apparently it does some magic that makes this fix important to not double freeing memory.

genaris (approved):

PR #255, opened 2023-10-20
fix leading zero bug

### LINKED ISSUE #252: Verifying a proof containing a value of "0" fails
When verifying a proof containing an attribute with the value "0" it will fail on the verification of the proof with the error:

```
AnoncredsError: Proof rejected: Encoded Values for "name" are different in RequestedProof "" and CryptoProof "0"
    at NodeJSAnoncreds.handleError (/Users/timo/Developer/Playground/anoncreds-0-test/node_modules/@hyperledger/anoncreds-nodejs/src/NodeJSAnoncreds.ts:51:11)
    at NodeJSAnoncreds.verifyPresentation (/Users/timo/Developer/Playground/anoncreds-0-test/node_modules/@hyperledger/anoncreds-nodejs/src/NodeJSAnoncreds.ts:464:10)
    at Presentation.verify (/Users/timo/Developer/Playground/anoncreds-0-test/node_modules/@hyperledger/anoncreds-shared/src/api/Presentation.ts:147:28)
    at runAnonCredsRs (/Users/timo/Developer/Playground/anoncreds-0-test/src/index.ts:206:31)
    at Object.<anonymous> (/Users/timo/Developer/Playground/anoncreds-0-test/src/index.ts:215:1)
    at Module._compile (node:internal/modules/cjs/loader:1254:14)
    at Module.m._compile (/Users/timo/Developer/Playground/anoncreds-0-test/node_modules/ts-node/src/index.ts:1618:23)
    at Module._extensions..js (node:internal/modules/cjs/loader:1308:10)
    at Object.require.extensions.<computed> [as .ts] (/Users/timo/Developer/Playground/anoncreds-0-test/node_modules/ts-node/src/index.ts:1621:12)
    at Module.load (node:internal/modules/cjs/loader:1117:32) {
  code: 7,
  extra: undefined
}
```

I've created the following script (using the JS wrapper) that reproduces the error:

<details>
<summary>JS reproduction</summary>

```ts
import {
  CredentialDefinition,
  CredentialOffer,
  CredentialRequest,
  LinkSecret,
  Presentation,
  PresentationRequest,
  Credential,
  Schema,
  anoncreds,
} from "@hyperledger/anoncreds-nodejs";


async function runAnonCredsRs() {
  const nonce = anoncreds.generateNonce();

  const presentationRequest = PresentationRequest.fromJson({
    nonce,
    name: "pres_req_1",
    version: "0.1",
    requested_attributes: {
      attr1_referent: {
        name: "name",
        restrictions: [{ cred_def_id: "mock:uri" }],
      },
    },
    requested_predicates: {},
  });

  const schema = Schema.create({
    name: "schema-1",
    issuerId: "mock:uri",
    version: "1",
    attributeNames: ["name"],
  });

  const {
    credentialDefinition,
    keyCorrectnessProof,
    credentialDefinitionPrivate,
  } = CredentialDefinition.create({
    schemaId: "mock:uri",
    issuerId: "mock:uri",
    schema,
    signatureType: "CL",
    tag: "TAG",
  });

  const credentialOffer = CredentialOffer.create({
    schemaId: "mock:uri",
    credentialDefinitionId: "mock:uri",
    keyCorrectnessProof,
  });

  const linkSecret = LinkSecret.create();
  const linkSecretId = "link secret id";

  const { credentialRequestMetadata, credentialRequest } =
    CredentialRequest.create({
      entropy: "entropy",
      credentialDefinition,
      linkSecret,
      linkSecretId,
      credentialOffer,
    });

  const credential = Credential.create({
    credentialDefinition,
    credentialDefinitionPrivate,
    credentialOffer,
    credentialRequest,
    attributeRawValues: { name: "0" },
  });

  const credentialReceived = credential.process({
    credentialDefinition,
    credentialRequestMetadata,
    linkSecret,
  });

  const presentation = Presentation.create({
    presentationRequest,
    credentials: [
      {
        credential: credentialReceived,
      },
    ],
    credentialDefinitions: { "mock:uri": credentialDefinition },
    credentialsProve: [
      {
        entryIndex: 0,
        isPredicate: false,
        referent: "attr1_referent",
        reveal: true,
      },
    ],
    linkSecret,
    schemas: { "mock:uri": schema },
    selfAttest: {},
  });

  const verify = presentation.verify({
    presentationRequest,
    schemas: { ["mock:uri"]: schema },
    credentialDefinitions: { ["mock:uri"]: credentialDefinition },
  });

  console.log(verify);
}

runAnonCredsRs();
```

</details>

If the value is changed to e.g. "10" in the script above, it will succeed.

I've also re-created the same test with the Indy SDK to see if this is behaviour that was inherited from that library. The same flow works fine with the Indy SDK

<details>
<summary>Indy SDK flow</summary>

```ts
import * as indySdk from "indy-sdk";

indySdk.setDefaultLogger("trace");

// @ts-ignore
indySdk.setRuntimeConfig({ collect_backtrace: true });

async function runIndySdk() {
  try {
    const id = "test-" + Math.random();
    await indySdk.createWallet({ id }, { key: "test" });
    const wh = await indySdk.openWallet({ id }, { key: "test" });

    const [did, verkey] = await indySdk.createAndStoreMyDid(wh, {});

    const schema = await indySdk.issuerCreateSchema(did, "schema-1", "1", [
      "name",
    ]);

    const [credDefId, credDefJson] =
      await indySdk.issuerCreateAndStoreCredentialDef(
        wh,
        did,
        schema[1],
        "TAG",
        "CL",
        { support_revocation: false }
      );

    const nonce = await indySdk.generateNonce();

    const credOffer = await indySdk.issuerCreateCredentialOffer(wh, credDefId);

    const masterSecretId = await indySdk.proverCreateMasterSecret(wh, "test");
    const [credReq, credReqMetadata] = await indySdk.proverCreateCredentialReq(
      wh,
      did,
      credOffer,
      credDefJson,
      masterSecretId
    );

    const cred = await indySdk.issuerCreateCredential(
      wh,
      credOffer,
      credReq,
      { name: { encoded: "0", raw: "0" } },
      null,
      -1
    );

    const pr = {
      nonce,
      name: "pres_req_1",
      version: "0.1",
      requested_attributes: {
        attr_ref_1: {
          name: "name",
          restrictions: [{ cred_def_id: credDefId }],
        },
      },
      requested_predicates: {},
    };

    const credId = await indySdk.proverStoreCredential(
      wh,
      null,
      credReqMetadata,
      cred[0],
      credDefJson,
      null
    );
    const proof = await indySdk.proverCreateProof(
      wh,
      pr,
      {
        requested_attributes: {
          attr_ref_1: {
            cred_id: credId,
            revealed: true,
          },
        },
        requested_predicates: {},
        self_attested_attributes: {},
      },
      masterSecretId,
      { [schema[0]]: schema[1] },
      { [credDefId]: credDefJson },
      {}
    );

    const valid = await indySdk.verifierVerifyProof(
      pr,
      proof,
      { [schema[0]]: schema[1] },
      { [credDefId]: credDefJson },
      {},
      {}
    );
    console.log(valid);
  } catch (error) {
    console.error(error);
  }
}

runIndySdk()
```

</details>

I haven't checked if this is also the case using credx, but curious to know if someone (@berendsliedrecht @andrewwhitehead ) knows whats going on here?

It should be supported it seems to include a value of 0? It seems to work fine with a predicate. (so requesting that the value is e.g. >= 0).

berendsliedrecht:
closes #252 

Minor bug where a value contains a "0" without having any digits behind it. 

Updated the regex to check for leading zeroes which must be followed by another digit. If this is the case, we trim the zeroes and return the leftover. If it is something like `0011212abab` it does not strip the zeroes.

https://github.com/hyperledger/aries-rfcs/tree/main/features/0592-indy-attachments#notes-on-encoding-claims

### DIFF
--- a/src/services/verifier.rs
+++ b/src/services/verifier.rs
@@ -450,15 +450,19 @@ fn verify_revealed_attribute_values(
     Ok(())
 }
 
+fn normalize_encoded_attr(attr: &str) -> String {
+    attr.parse::<i32>()
+        .map(|a| a.to_string())
+        .unwrap_or(attr.to_owned())
+}
+
 fn verify_revealed_attribute_value(
     attr_name: &str,
     proof: &Presentation,
     attr_info: &RevealedAttributeInfo,
 ) -> Result<()> {
-    let reveal_attr_encoded = attr_info.encoded.to_string();
-    let reveal_attr_encoded = Regex::new("^0*")
-        .unwrap()
-        .replace_all(&reveal_attr_encoded, "");
+    let reveal_attr_encoded = normalize_encoded_attr(&attr_info.encoded);
+
     let sub_proof_index = attr_info.sub_proof_index as usize;
 
     let crypto_proof_encoded = proof
@@ -1192,5 +1196,17 @@ mod tests {
             },
         );
         res
+    }
+
+    #[test]
+    fn format_attribute() {
+        assert_eq!(normalize_encoded_attr(""), "");
+        assert_eq!(normalize_encoded_attr("abc"), "abc");
+        assert_eq!(normalize_encoded_attr("0"), "0");
+        assert_eq!(normalize_encoded_attr("01"), "1");
+        assert_eq!(normalize_encoded_attr("01.0"), "01.0");
+        assert_eq!(normalize_encoded_attr("0abc"), "0abc");
+        assert_eq!(normalize_encoded_attr("-100"), "-100");
+        assert_eq!(normalize_encoded_attr("-0100"), "-100");
     }
 }


TimoGlastra (approved):
LGTM, but probably good if andrew also takes a look

src/services/verifier.rs
```diff
@@ -455,10 +455,15 @@ fn verify_revealed_attribute_value(
...
 ) -> Result<()> {
-    let reveal_attr_encoded = attr_info.encoded.to_string();
-    let reveal_attr_encoded = Regex::new("^0*")
-        .unwrap()
-        .replace_all(&reveal_attr_encoded, "");
+    let leading_zero_regex = Regex::new("^0[0-9]+").unwrap();
+    let encoded_info = &attr_info.encoded;
+
+    let reveal_attr_encoded = if leading_zero_regex.captures(encoded_info).is_some() {
+        encoded_info.trim_start_matches('0')
```
berendsliedrecht, line 462: I could not use the captured group sadly as it captures the entire number, seemes like an insignificant performance decrease to me.

TimoGlastra: You could create a capture group for this I think?

Not sure if it's better for performance, or if it's better at all. But this would work I think:

```
^(0+)[1-9]([0-9]+)?$
```

andrewwhitehead (approved):
I'll need to add a similar fix to credx. TBH I wasn't sure why it was trimming zeroes there when I came across it.

Related: https://github.com/hyperledger/indy-shared-rs/pull/43

src/services/verifier.rs
andrewwhitehead: Actually the regex probably should use $ to avoid stripping leading zeros on "01.0" for instance

berendsliedrecht: Moved from the regex to `String::parse<i32>` as it changes `-0100` to `-100`.

src/services/verifier.rs
```diff
@@ -1193,4 +1196,16 @@ mod tests {
...
+    #[test]
+    fn format_attribute() {
+        assert_eq!(normalize_encoded_attr(""), "");
+        assert_eq!(normalize_encoded_attr("abc"), "abc");
+        assert_eq!(normalize_encoded_attr("0"), "0");
+        assert_eq!(normalize_encoded_attr("01"), "1");
+        assert_eq!(normalize_encoded_attr("01.0"), "01.0");
+        assert_eq!(normalize_encoded_attr("0abc"), "0abc");
+        assert_eq!(normalize_encoded_attr("-100"), "-100");
+        assert_eq!(normalize_encoded_attr("-0100"), "-100");
```
TimoGlastra, line 1209: Would you expect this to have the leading 0 trimmed?

We need to make sure the spec aligns with this behavior

berendsliedrecht: The spec mentions  `convert any string integer (e.g. "1234") to be a 32-bit integer (e.g. 1234)`. Since integers can be negative, I thought this was the correct way. Maybe @swcurran can offer some clarification on this?

TimoGlastra: That makes sense yes 👍

PR #257, opened 2023-10-26
do not write to .npmrc for release

berendsliedrecht:
Signed-off-by: Berend Sliedrecht <blu3beri@proton.me>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -208,6 +208,7 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: 18.x
+          registry-url: "https://registry.npmjs.org/"
 
       - uses: pnpm/action-setup@v2
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -239,20 +240,14 @@ jobs:
           LIB_ANONCREDS_PATH: ../../
         run: pnpm test
 
-      - name: Set NPM config
```
TimoGlastra, line 242: This was added for a reason, let's hope it will work with the changes (also the always-auth seems to be removed)

TimoGlastra (approved):

.github/workflows/build.yml
berendsliedrecht: Ah yeah good catch. If that causes an issue I will add it back to the `.npmrc`.

PR #261, opened 2023-10-27
update js version

berendsliedrecht:
Signed-off-by: Berend Sliedrecht <blu3beri@proton.me>


TimoGlastra (approved):
For dev versions we usually keep the dev versions equal between all wrappers. This makes it a bit more easy to keep everything in sync IMO.

PR #266, opened 2023-11-02
Design: AnonCreds Credentials using the W3C Standard

Artemkaaas:
This PR contains the very first **DRAFT** working version of the design documents connected to adoption of AnonCreds Verifiable Credentials and Presentations for the W3C Verifiable Credentials Data Model Standard. 
The main goal of this PR is starting the discussion on the proposed format and implementation path / API.

Currently, the PR contains two main documents:
- `spec.md` - document describing how AnonCreds credentials and presentations should be represented in the W3C Data Model  
   - this document should not be merged into this repository
   - It should be moved and included into AnonCreds specification (I believe)
   - I put it here to have a single place for discussions
- `w3c-representation.md` -  document describing proposed `anoncreds-rs` API changes and implementation path.


docs/design/w3c/w3c-representation.md
```diff
@@ -0,0 +1,382 @@
...
+- easier deprecation of indy styled credentials and APIs
+- presentation conversion methods are not needed anymore in this case
+  - only credential conversion method to do migration for previously issued credentials
+
+```rust
+/// Create Credential Offer according to the AnonCreds specification
+/// It can be either Indy styled or W3C adopted depending on the answer for Q1 
+/// If Indy styled credential to be used, it should indicate that credential in W3C AnonCreds form will be issued as the result.
+///
+/// TO DISCUSS: Should we convert/adopt Credential Offer for W3C form?
```
TimoGlastra, line 197: Curious what a W3C form offer looks like.

For AFJ we were intending on superseding RFC 0593, supporting non linked-data signatures, multipe proof types and custom metadata so we can issue AnonCreds credentials (e.g. include cred_def_id and key_proof, etc.. in the proof, as well as have custom properties for the request)

TimoGlastra: As the offer is very dependant on which transport you're using, I don't think AnonCreds spec/lib can define a w3c format.

E.g. the OpenID4VCI has a different way to issue W3C credentials than Aries RFC 0593. The end result is the same: you have a W3C credential with a proof, but the exchange and communication is different.

TimoGlastra (commented):
Nice, this looks good. I've left some comments and questions, also to get a better understanding of how the result of this work can be integrated into AFJ (and ACA-Py)

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,530 @@
...
+schema in order to include the information about Indy related definitions to credential and looks the following:
+
+```
+{
+  ... 
+  "credentialSchema": {
+    "type": "AnonCredsDefinition",
+    "definition": "did:sov:3avoBCqDMFHFaKUHug9s8W:3:CL:13:default",
+    "schema": "did:sov:3avoBCqDMFHFaKUHug9s8W:2:fabername:0.1.0",
+    "encoding": "auto"
```
berendsliedrecht, line 169: Would be good to discuss if we would ever support another encoding scheme. If that is not the case, we can just omit this value.

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,530 @@
...
+      at [Aries RFC 0592 Indy Attachments section](https://github.com/hyperledger/aries-rfcs/tree/main/features/0592-indy-attachments#encoding-claims)
+      to generate encoded values under the hood during the signature generation and proof verification.
+
+#### Issuer
+
+W3C [Issuer](https://www.w3.org/TR/vc-data-model/#issuer) section requires including of `issuer` property to express the
+issuer of a verifiable credential.
+
+In the case of W3C AnonCreds credentials, the `issuer` attribute should be represented as a
+resolvable [DID URL](https://w3c-ccg.github.io/did-resolution/) having either `indy` or `sov` DID method and looks the
```
berendsliedrecht, line 191: I don't think we should only allow `indy` and `sov`, right?

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,530 @@
...
+  ...
+}
+```
+
+#### Issuance Date
+
+W3C [Issuance Date](https://www.w3.org/TR/vc-data-model/#issuance-date) section requires including of `issuanceDate`
+property to express the date and time when a credential becomes valid.
+
+In the case of W3C AnonCreds credentials, for the `issuanceDate` attribute recommended setting of a random time of the
```
berendsliedrecht, line 207: Any specific reason for a random time of the day? Couldn't we just say the exact time or set it to 00:00?

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,530 @@
...
+* `signature` - credential signature
+    * signature received by building the following object from indy styled credential:
+      ```
+      {
+        "signature": {..}, 
+        "signature_correctness_proof": {..}
+      }
+      ```
+    * encoded
+      as [base64 attachment](https://github.com/hyperledger/aries-rfcs/tree/main/concepts/0017-attachments#base64url)
```
berendsliedrecht, line 259: base64-url-no-pad I assume instead of base64?

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,530 @@
...
+
+
+* **TO DISCUSS**: Signature/Proof encoding: Which approach to use for encoding?
```
berendsliedrecht, line 262: I think we should really see the size difference here and how much it would add to an implementer to support this custom compact encoding. 

If the size difference is negligible or it is rather complex, I'd rather sick with base64url. If it is significant or trivial to implement the compact would be my preference.

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,530 @@
...
+
+##### Data Integrity proof
+
+In order to better conform to the W3C specification AnonCreds based credential also requires including
+of [Data Integrity Proof](https://www.w3.org/TR/vc-data-model/#data-integrity-proofs) which must be generated using one
+of NIST-approved algorithms (RSA, ECDSA, EdDSA).
+
+Including Data Integrity proof allows to use verifiable credential without access to a Ledger.
+
+#### Expiration
```
berendsliedrecht, line 274: I don't we can map this on a specification level between an anoncreds credential and w3c credential as anoncreds does not have this concept. In that case, I'd say that we should be strict that the anoncreds credential will not work with the `expirationDate` property.

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,530 @@
...
+The only difference is the value of the `proof` property.
+
+In the case of W3C AnonCreds presentations, the `proof` attribute uses defined `AnonCredsPresentationProof2022`
+type pointing to the difference in a presentation structure and looks the following:
+
+```
+  "proof": {
+    "type": "AnonCredsPresentationProof2022",
+    "credential": {
+      "mapping": {
```
berendsliedrecht, line 431: Any specific reason to nest this under a `mapping` key?

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,530 @@
...
+    * encoded
+      as [base64 attachment](https://github.com/hyperledger/aries-rfcs/tree/main/concepts/0017-attachments#base64url).
+* `mapping` - date requested in the proof request
+    * data: attribute name and reference in the proof request
+        * `revealedAttributes` - list of requested attributes revealed using the credential
+        * `unrevealedAttributes` - list of requested attributes presented in the credential but left unrevealed
+        * `requestedPredicates` - list of predicates resolved using the credential
+
+
+* **TO DISCUSS**: Should we remove `mapping` completely or move under encoded `proofValue`?
```
berendsliedrecht, line 477: Ah, did not read this yet. I'd say we omit it and if we need it for validation we can add it back before validation.

docs/design/w3c/w3c-representation.md
```diff
@@ -0,0 +1,382 @@
...
+
+This design document proposes extending of `anoncreds-rs` library to add support for AnonCreds W3C representation of verifiable credential and presentation described in the [document]().
+
+### Goals and ideas
+
+* Use Indy styled credentials to generate W3C AnonCreds credentials and presentations
+  * Credentials conversion:
+    * Convert Indy styled AnonCreds credentials into W3C form
+    * Convert W3C styled AnonCreds credentials into Indy form
+  * Presentation conversion (Optional):
```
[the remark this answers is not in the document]
berendsliedrecht: I'd also say that this should be required.

docs/design/w3c/w3c-representation.md
```diff
@@ -0,0 +1,387 @@
...
+* Presentations: Support different formats (for example DIF) of Presentation Request
+
+### Question impacting the approach
+
+* Q1: Do we need conversion for intermediate messages to make them W3C compliant: Credential Offer, Credential Request?
+  * Q1.1: (Depends on answer for Q1) If no conversion: Do we want Credential Offer indicates what form of credential will be issued as the process execution result?
+* Q2: Do we want duplicate methods (like `sign` and `sign_w3c`) or only use single conversion method (like `credential.to_w3c`) doing extra step?
+* Q3: Should we care about the legacy credential migration and compatibility? Or achieve better following to the W3C standard with preserving power of CL signatures?
+  * example pros of dropping legacy support:
+    * different attributes types can be used in credentialSubject (support nested objects, array, other features)
```
berendsliedrecht, line 35: This would be a nice feature, but I think we will not be able to create a mapping anymore between anoncreds and w3c. Unless we encode the nested objects, but I am not the biggest fan of that.

berendsliedrecht: I'd also assume we can not do selective disclosures / predicates over those nested attributes, or can we?

docs/design/w3c/w3c-representation.md
```diff
@@ -0,0 +1,387 @@
...
+```
+
+#### Credential object methods
+
+```rust
+/// Get W3C credential payload to sign for making Identity Proof
+///
+/// # Params
+/// cred -                  object handle pointing to W3C styled AnonCreds credential
+/// signing_payload_p -     reference that will contain payload to sign for generation of data integrity proof
```
berendsliedrecht, line 109: I am really not too familiar with the data integrity proof, but I don't see the creation of the input that will be signed for a data integrity proof to be the responsibility of this library, right? Might be very wrong here, so please correct me if that is the case.

berendsliedrecht: Or will the CL sig be a data integrity proof?

berendsliedrecht (commented):
I am not very up to date with the intricate details of the w3c vc data model 1.1 and data integrity proof so if I mentioned something that does not make sense, feel free to ignore it.

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,530 @@
...
+      "created": "2021-11-13T18:19:39Z",
+      "verificationMethod": "did:sov:3avoBCqDMFHFaKUHug9s8W#key-1",
+      "proofPurpose": "assertionMethod",
+      "proofValue": "z58DAdFfa9SkqZMVPxAQpic7ndSayn1PzZs6ZjWp1CktyGesjuTSwRdoWhAfGFCF5bppETSTojQCrfFPP2oumHKtz"
+    }
+  ]
+}
+```
+
+#### Context
```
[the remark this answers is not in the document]
swcurran: My $0.02 — I think for the AnonCreds processing, there should not be any other checking beyond that the two required contexts are there.  Others could be added, but they would not be processed, per JSON-LD best practices.  A “general” JSON-LD processor should not dynamically load contexts, and AnonCreds fits that category.

I suppose the two required contexts could be hard-coded into the implementation — e.g. pre-loaded. I’m not sure what value that would give, as anything not included in those two would become whatever “undefined”.

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,545 @@
...
+  ...
+}
+```
+
+#### Identifiers
+
+W3C [Identifiers](https://www.w3.org/TR/vc-data-model/#identifiers) section defines an optional capability to assign
+some kind of identifier to the verifiable credential so that others can express statements about the same thing.
+
+In the case of W3C AnonCreds credentials, the `id` attribute is not recommended to set.
```
swcurran, line 83: For AnonCreds v1 — this work — this is the case.  In a future AnonCreds, optional metadata **could** be added to the AnonCreds data to identify an “id” attribute that could be used, but definitely not in this round, where we are not changing AnonCreds at all.

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,530 @@
...
+{
+  ... 
+  "credentialSubject": {
+    "name": "Alice Jones",
+  }
+  ...
+}
+```
+
+* **TO DISCUSS**: W3C data model allows attributes to be represented not only as key/value string pairs but also as
```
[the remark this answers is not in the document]
swcurran: Per the goal of this effort, any AnonCreds v1 credential can be represented as an AnonCreds in W3C VC format, but not the reverse — an arbitrary W3C VC format credential (notably, those with nested attributes and arrays) will not be supported in this version.

Longer term — beyond this immediate work — arrays are out, as they are not supported by CL Signatures, and they are extremely difficult to request/present.  Nested attributes could easily be supported using a dot-notation attribute list.

As always, those for whom array support is necessary can do so with an “rich attribute value” — e.g. use a data url to encode JSON into the array.  Can’t selectively disclose the array items, but data could be transported in the VC/VP.

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,545 @@
+## W3C Verifiable Credentials Representation
+
+This section describes how Indy-styled AnonCreds credentials can be represented in the form of W3C Verifiable
```
swcurran, line 3: Throughout this document, the term “Indy” should be removed, as AnonCreds are stand-alone and independent of Indy.  In most cases in this text, “AnonCreds” is suitable replacement. Obviously not in this case 😄.  I think we should use the terms “legacy AnonCreds format” and “W3C AnonCreds format” for the type representations. Where necessary “VC” and “VP” would be put in front of “format”.

docs/design/w3c/spec.md
swcurran: For AnonCreds 1.0, we have no plans to support any other encoding scheme and an update to the spec. in that area is extremely unlikely.

Converting to `”encoding": “auto"` is a change, in that we can take the encoding out of the hands of the issuer and do it in the library.  I’m a fan of that, but I think we should leave in the `encoding`, since it allows for us to stick to the current implementation — e.g., putting in what the issuer provided.

```
  ... 
  “encoding": {
    "name": “12341241535",
  }
  ...

```

Definitely worth a discussion — even to the point of in the legacy AnonCreds format allowing for an `”encoding": “auto”`option where the issuer-provided encoding is ignored and or/rejected if inaccurate.

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,545 @@
...
+    "type": "AnonCredsDefinition",
+    "definition": "did:sov:3avoBCqDMFHFaKUHug9s8W:3:CL:13:default",
+    "schema": "did:sov:3avoBCqDMFHFaKUHug9s8W:2:fabername:0.1.0",
+    "encoding": "auto"
+  },
+  ...
+}
+```
+
+* `schema` - id of Indy Schema
```
swcurran, line 175: In this section, replace the “Indy” references to be AnonCreds.

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,545 @@
...
+}
+```
+
+#### Proofs (Signatures)
+
+W3C [Proofs (Signatures)](https://www.w3.org/TR/vc-data-model/#proofs-signatures) section requires including of `proof`
+property to express confirmation of the credential's validity.
+
+According to the specification, one or many proof objects can be added to verifiable credentials.
+In the context of W3C AnonCreds credentials included at least two proof object entries: AnonCreds CL
```
swcurran, line 224: Wording has to be handled carefully here. The plan outlined here specifies an AnonCreds signature as a Data Integrity Proof.  A second, non-AnonCreds signature would also be a Data Integrity Proof — just one using a different signature scheme.  So the wording here is not quite right.

docs/design/w3c/spec.md
swcurran: I think the more complex encoding is fine, vs. straight serialized JSON encoding.  But agree that seeing the size difference would be the right thing to understand the tradeoff.

docs/design/w3c/spec.md
swcurran: Agree. As with the `credentialSubject: id`, we might sometime later add metadata such that the issuer can indicate that `attribute X contains the expiry date`, but that adds a bunch of uncertainty about date format, etc.  Not for now, I don’t think.

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,469 @@
...
+include a revocation registry id into the credential schema.
+
+#### Status
+
+W3C [Status](https://www.w3.org/TR/vc-data-model/#status) section defines an optional capability to include credential
+status information.
+
+Instead of including `credentialStatus` property we recommend using a standard indy credentials revocation approach and
+include a revocation registry id into the credential schema.
+
```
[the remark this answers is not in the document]
swcurran: I was hoping we can do that as well.  In theory, the same AnonCreds revocation registry can be used for both the AnonCreds case (using a ZKP non-revocation proof) and for the non-ZKP case, where the holder tells the verifier “here is the registry to use, my index in it”. In the latter case, the verifier can check with the source (whereever that is) to determine the revocation status.

If we can move on a better ZKP-based revocation scheme (such as the one @andrewwhitehead has proposed) this would be really useful to support.

docs/design/w3c/spec.md
swcurran: I like the idea of the VC that contains the mapping and the proofs (primary, predicates and non-revocation) for each credential. Makes it very clear the scope of the proofs are derived from a single credential. There would be one per source credential in the presentation.

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,545 @@
...
+* `mapping` - date requested in the proof request
+    * data: attribute name and reference in the proof request
+        * `revealedAttributes` - list of requested attributes revealed using the credential
+        * `unrevealedAttributes` - list of requested attributes presented in the credential but left unrevealed
+        * `requestedPredicates` - list of predicates resolved using the credential
+
+
+* **TO DISCUSS**: Should we remove `mapping` completely or move under encoded `proofValue`?
+    * Why `mapping` is bad: we make presentation tied to Indy styled Presentation Request
+    * Mapping is something old indy-fashioned required for validation (not signature verification) that proof matches to
```
swcurran, line 481: The mapping is necessary for the verifier to align with the presentation request.  I don’t see how we can drop it.

docs/design/w3c/w3c-representation.md
```diff
@@ -0,0 +1,457 @@
...
+### Goals and ideas
+
+* Use Indy styled credentials to generate W3C AnonCreds credentials and presentations
+    * Credentials conversion:
+        * Convert Indy styled AnonCreds credentials into W3C form
+        * Convert W3C styled AnonCreds credentials into Indy form
+    * Presentation conversion (Optional):
+        * Convert Indy styled AnonCreds presentation into W3C form
+        * Convert W3C styled AnonCreds presentation into Indy form
+* Extend W3C credentials:
```
swcurran, line 18: As noted earlier — careful with the wording here, as the W3C AnonCreds format credentials are Data Integrity Proofs.

docs/design/w3c/w3c-representation.md
```diff
@@ -0,0 +1,457 @@
...
+    * Ability to set Data Integrity proof signatures for generated W3C credential objects:
+        * W3C credentials may contain multiple signatures
+        * AnonCreds-Rs only generates/handle AnonCreds signatures
+
+#### Out of scope
+
+* Credentials: Verify validity of the Data Integrity proof signature
```
swcurran, line 25: Change to `Verify validity of non-AnonCreds Data Integrity proof signatures`

I agree that the intention is correct.

swcurran: Same as for the following...

docs/design/w3c/w3c-representation.md
```diff
@@ -0,0 +1,457 @@
...
+* Presentations: Create presentation using Data Integrity proof signature
+* Presentations: Verify validity of presentations using Data Integrity proof signatures
+* Presentations: Support different formats (for example DIF) of Presentation Request
+
+### Question impacting the approach
+
+* Q1: Do we need conversion for intermediate messages to make them W3C compliant: Credential Offer, Credential Request?
```
swcurran, line 32: In my opinion — no.  The only things that need to be W3C VC compliant are the VC and VP.

docs/design/w3c/w3c-representation.md
```diff
@@ -0,0 +1,457 @@
...
+    * Q1.1: (Depends on answer for Q1) If no conversion: Do we want Credential Offer indicates what form of credential
+      will be issued as the process execution result?
+    * Proposed answer: no conversion for the current phase
+* Q2: Do we want duplicate methods (like `sign` and `sign_w3c`) or only use single conversion method (
+  like `credential.to_w3c`) doing extra step?
+    * There are 6 methods in total. 4 of them we have to duplicate any way. Whether we want to
+      duplicate `create_offer`, `create_request` methods if we do not change their format.
+    * Proposed answer: yes - duplicate all methods
+* Q3: Are we still tied to Indy styled presentation request?
```
swcurran, line 41: I think we are tied to the existing AnonCreds Presentation Request.  We are not proposing an update from that in this effort.

swcurran: It would be nice for AnonCreds to support Presentation Exchange, but that is (IMHO) out of scope of this work.  We’ve talked about doing that in the past, and I would love to see it, likely using the same approach — transform a PEx to AnonCreds Presentation Request and back.  That said, I’ve done no research into whether that is possible.

docs/design/w3c/w3c-representation.md
```diff
@@ -0,0 +1,457 @@
...
+    * Can we make interface completely independent of Presentation Request? So any form can be handled on top level and
+      specific data passed to AnonCreds-rs.
+    * Proposed answer: Make AnonCreds-rs methods presentation agnostic.
+* Q4: Do we want to provide general interfaces doing signing and verification of Data Integrity proof signature?
```
swcurran, line 45: I assume by that you mean “Non-AnonCreds Data Integrity Proof signatures”.  If so, I’d say no.  The VC can be passed to another library if anyone is interested in that.

A flow might be — pass to AnonCreds for a “sign_w3c”, strip off the AnonCreds proof (if necessary), pass the result to another library to get another signature, re-insert the AnonCreds proof.

docs/design/w3c/w3c-representation.md
```diff
@@ -0,0 +1,457 @@
...
+    * Accept sign/verify callbacks into convert/create functions:
+        * Issue with setting multiple signatures
+        * Much easier just to expose methods to add signature proof itself
+    * Provide methods to put ready data into a credential
+    * Proposed answer: delegate to third party libraries using `anoncreds-rs`
+* Q5: Signature/Proof encoding: Which approach to use for encoding?
+    * Basic approach used in Aries attachments - Base 64 encoding of serialized
+      object (`base64(json(signature).as_bytes())`) ?
+    * Compact encoding implementing in Python PoC?
+    * Proposed answer: Start from basic approach?
```
swcurran, line 55: Do we have any idea of the size benefit?  Would a quick experiment help?  What did @andrewwhitehead do in the Python POC — https://github.com/andrewwhitehead/anoncreds-w3c-mapping?

docs/design/w3c/w3c-representation.md
```diff
@@ -0,0 +1,457 @@
...
+* Q6: Should we care about back way conversion of credential issued in W3C form?
```
swcurran, line 56: No 😄

docs/design/w3c/w3c-representation.md
swcurran: In OpenID4VCs, there is a request-response protocol for issuance. However, I’ve had two people tell me that the required Offer-Request-Issue protocol for AnonCreds issuance is possible using OpenID4VCs.

swcurran:
Great stuff in this.  We’ll be discussing this at the AnonCreds meeting on Monday Nov. 6, 7:00 Pacific / 16:00 Central Europe for those that can make it. I understand it is a holiday for some, so not everyone can make it, but we’ll do what we can.

Meeting info: https://wiki.hyperledger.org/display/ANONCREDS/2023-11-06+AnonCreds+Working+Group+Meeting

docs/design/w3c/w3c-representation.md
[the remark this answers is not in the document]
ashcherbakov: It's marked optional, because it may not be needed if we decide to have duplicate methods for w3c (see Q2) which is a recommended approach. So, instead of having `anoncreds_presentation_to_w3c`, the holder can use `anoncreds_w3c_create_presentation` (w3c analogue of `anoncreds_create_presentation`).
Whether the conversion method is needed for the Verifier will depend on how holder and verifier negotiate what version of the protocol (w3c based or legacy Anoncreds based) they will use.

docs/design/w3c/w3c-representation.md
```diff
@@ -0,0 +1,382 @@
...
+  * Q1.1: (Depends on answer for Q1) If no conversion: Do we want Credential Offer indicates what form of credential will be issued as the process execution result?
+* Q2: Do we want duplicate methods (like `sign` and `sign_w3c`) or only use single conversion method (like `credential.to_w3c`) doing extra step?
+* Q3: Should we care about the legacy credential migration and compatibility? Or achieve better following to the W3C standard with preserving power of CL signatures?
+  * example pros of dropping legacy support:
+    * different attributes types can be used in credentialSubject (support nested objects, array, other features)
+      * Need to decide on object encoding algorithm
+    * attributes mapping in presentation request can be simplified but with loosing ability to make back conversion
+* Q4: Are we still tied to Indy styled presentation request?
+  * Q4.1: (Depends on answer for Q3) If yes: Do we want Presentation Request indicates what form of presentation need to be created (`PresentationRequestV3`)?
+  * Q4.2: (Depends on answer for Q3) If no: What Presentation Request format need to be supported? DIF?
```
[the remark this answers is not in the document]
ashcherbakov: Yes, we also realised that for the current anoncreds-rs API it's out of scope, so that it can be made presentation agnostic.

docs/design/w3c/w3c-representation.md
ashcherbakov: I think it depends on the way how the holder and verifier negotiate about the format of presentation (on the protocol level?), especially in the legacy cases and for existing applications. What if the holder has only w3c credentials, but verifier uses an old implementation that can only work with a legacy AnonCreds format? In this case we should either raise Unsupported error, or let the holder convert the w3c VC/VP into legacy one to be understood by the verifier.

docs/design/w3c/w3c-representation.md
ashcherbakov: Please have a look at https://github.com/hyperledger/anoncreds-rs/blob/9e7763cd66c9a819ad7852f88b31912d853d81a1/docs/design/w3c/encoding.md#example

docs/design/w3c/w3c-representation.md
TimoGlastra: > I’ve had two people tell me that the required Offer-Request-Issue protocol for AnonCreds issuance is possible using OpenID4VCs.

Yes, but it's only possible if we define a custom credential format profile specific to AnonCreds. It can't leverage the existing JSON-LD format defined in the spec AFAIK. Although I'm not sure extra properties are not allowed, so it could be an extension to the JSON-LD format maybe (to allow to pass the AnonCreds offer/request data)

docs/design/w3c/spec.md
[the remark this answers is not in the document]
TimoGlastra: > arrays are out, as they are not supported by CL Signatures

I'm curious. What is the reason for using a different signing key for every attribute in an anoncreds credential? Because if the same signing key would be used for all array entries under a key it should be quite doable to supports arrays right?

docs/design/w3c/w3c-representation.md
TimoGlastra: I think this is definitely a very useful feature, as @ashcherbakov pointed out. In AFJ we intend to store all AnonCreds credentials going forward in the W3C format, and if doing an exchange that's not using the W3C format, we covert the credential back to it's AnonCreds format.

If not provided by this library, it would mean we have to implement that in AFJ

swcurran: Ah…sorry — I didn’t understand the context.  Absolutely there should be a two-way conversion — I had just assumed that would be necessary.  I was thinking you meant a non-AnonCreds W3C VC Format...

ashcherbakov:
A list of open questions (some with proposed answers) in one place (to be extended / updated): [link](https://github.com/hyperledger/anoncreds-rs/pull/266/files#diff-f0f0c92035decc44061ca415febaa763d3b4b86afc79572ba83f4d76d2a0f617R30)

docs/design/w3c/w3c-representation.md
ashcherbakov: Also, I think Artem's initial question was related not so much to a need of such conversion (it seems we all think it's needed and it's not hard to implement it), but rather to a discussion about support of hierarchical claims (nested objects) in addition to plain key-values. Conversion of hierarchical claims (nested objects) to a legacy format may be tricky.

docs/design/w3c/encoding.md
```diff
@@ -0,0 +1,113 @@
+### Encoding
+
+As all fields of AnonCreds credential signature and proof objects are big numbers, the straight object JSON serialization and representing as bytes is not effective.
```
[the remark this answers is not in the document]
andrewwhitehead: The encoding used in the PoC is just a placeholder really, although I did want to see how small the result could be. I think that CBOR would be a better option, either with a set of registered object keys or using arrays everywhere as below.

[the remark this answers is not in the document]
Artemkaaas: At the yesterday's Aries WG call, we agreed to make the first implementation with the default Aries attachment encoding to deliver library faster, and as the next step experiment with different encoding algorithms (custom from PoC, msgpack, cbor, protobuf) and use one of them.

docs/design/w3c/encoding.md
```diff
@@ -0,0 +1,113 @@
...
+Instead, we propose using of an alternative algorithm, providing more compact representation, consisting of the following steps:
+
+Also, due to the fact that fields order matters during serialization/deserialization process, encoding must be applied to attributes in alphabetic order.
```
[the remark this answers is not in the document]
Artemkaaas: In fact, encoding defined for specific data types with the known limited field name set. 
No custom fields expected I suppose.

docs/design/w3c/spec.md
Artemkaaas: Right. Basically, I just implied the reference to the RFC section here. Not exact format.

docs/design/w3c/w3c-representation.md
Artemkaaas: You are right. I'll remove this method. It will be the responsibility of 3td party library to create/verify non-AnonCreds identity proof signature.

docs/design/w3c/spec.md
Artemkaaas: @swcurran  Please have a look at this pseudo code: https://github.com/hyperledger/anoncreds-rs/pull/266/files#diff-f0f0c92035decc44061ca415febaa763d3b4b86afc79572ba83f4d76d2a0f617R668

swcurran: I'm not sure what I'm looking at there -- the link takes me to all the files in the commit. How about we cover this on Monday -- you can show what you mean and we can discuss what happens if we don't include it.

docs/design/w3c/w3c-representation.md
swcurran: And that is definitely out of scope for this.  Only current style "list of attribute" AnonCreds credentials are in scope.

docs/design/w3c/w3c-representation.md
```diff
@@ -0,0 +1,786 @@
...
+        * During the `proofValue` crypto verification we can parse the phrase and restore predicate attributes
+    * Put predicate as object into `credentialSubject`
+        ```
+          "credentialSubject": {
+            ...
+            "birthdate": {
+              "type": "Predicate",
+              "p_type": "<", 
+              "value": "20041012"
+            }
```
andrewwhitehead, line 111: I think this syntax could work, but implementations should handle an array value here in the case of multiple predicates on the same attribute. Or both a value reveal and a predicate on the same attribute, for example:

```json
            "birthdate": [
              "20040101",
              {
                "type": "Predicate",
                "p_type": "<", 
                "value": "20041012"
              }
            ],
```

Artemkaaas: This case makes sense: `multiple predicates on the same attribute`
But this one is not valid and will fail on the `cl` library level: `both a value reveal and a predicate on the same attribute`

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,467 @@
...
+```
+  "issuer": "did:sov:3avoBCqDMFHFaKUHug9s8W",
+```
+
+#### Issuance Date
+
+W3C [Issuance Date](https://www.w3.org/TR/vc-data-model/#issuance-date) section requires including of `issuanceDate`
+property to express the date and time when a credential becomes valid.
+
+In the context of W3C AnonCreds credentials, in order to avoid correlation, recommended setting of a random time of the
```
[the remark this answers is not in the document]
Artemkaaas: At the last Aries AnonCreds community call we decided just to use the current time when credential was issued/converted

docs/design/w3c/w3c-representation.md
Artemkaaas: Discussed during the Aries Announced community call. 
Decided to start with the base base64-url-safe encoding and reach other encoding algorithms as the next step.

docs/design/w3c/spec.md
```diff
@@ -0,0 +1,469 @@
...
+      the request itself on the verifier side
+    * For doing crypto `proofValue` verification we only need the names of revealed attributes and predicated (with
+      type)
+
+* TO DISCUSS 2: Should we derive an attribute from a predicate and put them into credentialSubject like it demonstrated
+  in the [specification](https://www.w3.org/TR/vc-data-model/#presentations-using-derived-credentials)
+    * Example:
+        * For Predicate: `{"name": "birthdate", "p_type": "<", "value":"20041012"}`
+        * Derived attribute: `{"birthdate": "birthdate less 20041012"}`
+    * During the `proofValue` crypto verification we can parse the phrase and restore predicate attributes
```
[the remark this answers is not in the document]
Artemkaaas: Discussed during  the Aries Announced community call. 
Decided to use object representation:
```
"age":{
          "type":"AnonCredsPredicate",
          "p_type":">=",
          "p_value":18
}
```

Artemkaaas:
I opened [PR](https://github.com/hyperledger/anoncreds-spec/pull/185) in [anoncreds-spec](https://github.com/hyperledger/anoncreds-spec)   repository which add W3C section into AnonCreds credentials specification. 

I'm going to exclude `context.json` and `spec.md` files from this PR (`anoncreds-rs`) as these files included into mentioned above PR to `anoncreds-spec` repository

Artemkaaas:
https://github.com/hyperledger/anoncreds-rs/pull/266#issuecomment-1822911739

@vortex-hue what do you mean? The flow itself keeps the same as before ([Spec](https://hyperledger.github.io/anoncreds-spec/#anoncreds-methods)). Now you have a set of parallel methods to get data in different format.

docs/design/w3c/w3c-representation.md
```diff
@@ -0,0 +1,886 @@
...
+The reasons for adding duplication methods:
+
+- avoid breaking changes in the existing API
+    - for example if we want Credential Offer pointing to the form of a credential to be issued
+- clear separation between flows
+    - if a flow targeting issuing of W3C Credential the specific set of function to be used
+- avoid the situation when function result value may be in different forms
+    - example:
+        - issuer creates offer in legacy form but with resulting credential format indication (legacy or w3c )
+        - as the flow execution result, create credential function returns credential either in w3c or legacy form
```
gvelez17, line 340: is it possible they will want both?

Artemkaaas: Yes. Both versions can be used depending on the higher level issuance protocol version

swcurran (approved):

swcurran:
Anyone know why a test is failing here?  Re-ran the failed test and got it in Ubuntu GHA (Windows the previous time).  We can merge this documentation only PR with this, but didn't want the error to be ignored.

Artemkaaas:
@swcurran Should I clean the section with questions before merge?

swcurran:
> @swcurran Should I clean the section with questions before merge?

It would be appreciated. Nice to have as well changing the “in the future” items to be now. For example, the for example, the first line that says (more or less) “Currently we don’t support W3C VC Format”. It should say something like “Here is how we support W3C VC Format…”.

Not crucial, but this will be long lived docs, so clean up is good.

andrewwhitehead (approved):

PR #273, opened 2023-11-14
AnonCreds Credentials using the W3C Standard - wrappers

Artemkaaas:
* Updated Python wrapper to expose new methods
* Update NodeJS and React Native wrappers to expose new methods


src/data_types/credential.rs
```diff
@@ -73,14 +75,71 @@ impl Validatable for Credential {
...
+#[cfg(feature = "zeroize")]
+impl Zeroize for RawCredentialValues {
+    fn zeroize(&mut self) {
+        for attr in self.0.values_mut() {
+            attr.zeroize();
+        }
+    }
+}
+
+impl Validatable for RawCredentialValues {
```
gvelez17, line 104: just curious what is the reason for change from Short to Raw is this because we need richer data for the flex creds?

Artemkaaas: I just considered name `short` not clear and confusing. can be reverted

wrappers/python/anoncreds/bindings.py
```diff
@@ -964,3 +1001,257 @@ def create_or_update_revocation_state(
...
+
+
+def create_w3c_credential(
+    cred_def: ObjectHandle,
+    cred_def_private: ObjectHandle,
+    cred_offer: ObjectHandle,
+    cred_request: ObjectHandle,
+    attr_raw_values: Mapping[str, str],
+    revocation_config: Optional[RevocationConfig],
+    encoding: Optional[str],
```
gvelez17, line 1050: @Artemkaaas why change the signature of the attributes?  the original create_credential is

```
 attr_raw_values: Mapping[str, str],
    attr_enc_values: Optional[Mapping[str, str]],
    revocation_config: Optional[RevocationConfig],
    ```
    
    maybe good to keep the signature the same if we can?

Comment (Artemkaaas) [reply wrappers/python/anoncreds/bindings.py:1050]: It was decided to move encoding of credential values to the library

File: wrappers/javascript/packages/anoncreds-nodejs/src/NodeJSAnoncreds.ts [hunk 3]
```diff
@@ -102,6 +102,16 @@ export class NodeJSAnoncreds implements Anoncreds {
     return handleReturnPointer<string>(ret)
   }
 
+  public w3cCredentialGetAttribute(options: { objectHandle: ObjectHandle; name: string }) {
```
TimoGlastra, line 105: Is there a reason why there is a separate method for w3c? Getting an attribute is the same for w3c and non w3c credentials right?

wrappers/javascript/packages/anoncreds-nodejs/src/NodeJSAnoncreds.ts
```diff
@@ -617,6 +627,380 @@ export class NodeJSAnoncreds implements Anoncreds {
     return new ObjectHandle(handleReturnPointer<number>(ret))
   }
 
+  public createW3CCredentialOffer(options: {
```
TimoGlastra, line 630:
```suggestion
  public createW3cCredentialOffer(options: {
```

To not have confusion in casing, we always only capitalize the first letter of a word (and W3c is one word in this case)

wrappers/javascript/packages/anoncreds-nodejs/src/NodeJSAnoncreds.ts
```diff
@@ -617,6 +627,380 @@ export class NodeJSAnoncreds implements Anoncreds {
...
+    schemaId: string
+    credentialDefinitionId: string
+    keyCorrectnessProof: ObjectHandle
+  }): ObjectHandle {
+    const { schemaId, credentialDefinitionId, keyCorrectnessProof } = serializeArguments(options)
+
+    const ret = allocatePointer()
+    this.nativeAnoncreds.anoncreds_create_w3c_credential_offer(
```
TimoGlastra, line 638: As I mentioned in another comment before, I do't think we need a w3c credential offer method, as the offer is not any different based on whether it's a w3c or anoncreds offer? 

The only difference is the end-result: whether the credential is encoded as AnonCreds or W3C credential

wrappers/javascript/packages/anoncreds-nodejs/src/NodeJSAnoncreds.ts
```diff
@@ -617,6 +627,380 @@ export class NodeJSAnoncreds implements Anoncreds {
...
+      credentialDefinitionId,
+      keyCorrectnessProof,
+      ret
+    )
+    this.handleError()
+
+    return new ObjectHandle(handleReturnPointer<number>(ret))
+  }
+
+  public createW3CCredentialRequest(options: {
```
TimoGlastra, line 649: Same here: `W3c`

and also, I don't see a reason for a w3c credential request, only the end result is different I think?

wrappers/javascript/packages/anoncreds-nodejs/src/NodeJSAnoncreds.ts
```diff
@@ -617,6 +627,380 @@ export class NodeJSAnoncreds implements Anoncreds {
...
+
+    const credentialEntryList = CredentialEntryListStruct({
+      count: credentialEntries.length,
+      data: credentialEntries as unknown as TypedArray<
+        StructObject<{
+          credential: number
+          timestamp: number
+          rev_state: number
+        }>
+      >
```
TimoGlastra, line 783: I think there's a lot of code here for w3c methods that is basically the same except for the end result. Do you think we can reuse some code here? I think this can lead to inconsistencies in functionality

wrappers/javascript/packages/anoncreds-nodejs/src/NodeJSAnoncreds.ts
```diff
@@ -617,6 +627,380 @@ export class NodeJSAnoncreds implements Anoncreds {
...
+
+    const ret = allocatePointer()
+
+    this.nativeAnoncreds.anoncreds_w3c_credential_add_non_anoncreds_integrity_proof(objectHandle, proof, ret)
+    this.handleError()
+
+    return new ObjectHandle(handleReturnPointer<number>(ret))
+  }
+
+  public w3cCredentialAddNonAnonCredsIntegrityProof(options: {
```
TimoGlastra, line 946: As mentioned in other PR, I don't think this should be handled by the AnonCreds library

wrappers/javascript/packages/anoncreds-nodejs/src/NodeJSAnoncreds.ts
```diff
@@ -617,6 +627,380 @@ export class NodeJSAnoncreds implements Anoncreds {
...
+
+    const ret = allocatePointer()
+
+    this.nativeAnoncreds.anoncreds_credential_from_w3c(objectHandle, ret)
+    this.handleError()
+
+    return new ObjectHandle(handleReturnPointer<number>(ret))
+  }
+
+  public w3cCredentialAddNonAnoncredsIntegrityProof(options: {
```
TimoGlastra, line 932: This method is basically duplicate from the method below

wrappers/javascript/packages/anoncreds-nodejs/src/NodeJSAnoncreds.ts
```diff
@@ -617,6 +627,380 @@ export class NodeJSAnoncreds implements Anoncreds {
...
+  public w3cCredentialSetSubjectId(options: { objectHandle: ObjectHandle; id: string }): ObjectHandle {
+    const { objectHandle, id } = serializeArguments(options)
+
+    const ret = allocatePointer()
+
+    this.nativeAnoncreds.anoncreds_w3c_credential_set_subject_id(objectHandle, id, ret)
+    this.handleError()
+
+    return new ObjectHandle(handleReturnPointer<number>(ret))
+  }
```
TimoGlastra, line 980: This feels like it's not really useful in the context of AnonCreds credentials, as you can't use them. Honestly, I'd like to keep it simple and not expose these kind of methods, but just have a simple method like you have with the normal anoncreds encoding (sign based on the attribute map you pass).

That way you can get an anoncreds credential in w3c encoding, if you want to add other contexts, or do special things for using it with other proof types, that should be handled outside the anoncreds library. It has nothing to do with AnonCreds. IMO this library should do the minimal amount of processing it can, and all custom ways should be handled by the frameworks consuming this library.

wrappers/javascript/packages/anoncreds-nodejs/src/NodeJSAnoncreds.ts
```diff
@@ -706,6 +1090,22 @@ export class NodeJSAnoncreds implements Anoncreds {
     return this.objectFromJson(this.nativeAnoncreds.anoncreds_key_correctness_proof_from_json, options)
   }
 
+  public w3cCredentialOfferFromJson(options: { json: string }): ObjectHandle {
+    return this.objectFromJson(this.nativeAnoncreds.anoncreds_credential_offer_from_json, options)
+  }
+
+  public w3cCredentialRequestFromJson(options: { json: string }): ObjectHandle {
+    return this.objectFromJson(this.nativeAnoncreds.anoncreds_credential_request_from_json, options)
+  }
```
TimoGlastra, line 1099: As these just call `anoncreds_credential_offer_from_json` I think we shouldn't add these new methods. Duplicating all methods doesn't make sense to me, and I think we should keep the duplication of methods only to the methods where it's actually needed

wrappers/javascript/packages/anoncreds-nodejs/tests/api.test.ts
```diff
@@ -540,3 +544,326 @@ test('create and verify presentation passing only JSON objects as parameters', (
...
+      credentialDefinition,
+      revocationRegistryDefinitionId: 'mock:uri',
+      revocationRegistryDefinition,
+      revocationRegistryDefinitionPrivate,
+      issuerId: 'mock:uri',
+      issuanceByDefault: true,
+      timestamp: timeCreateRevStatusList
+    })
+
+    const credentialOffer = W3CCredentialOffer.create({
```
TimoGlastra, line 611: This can just be `CredentialOffer.create` right?

wrappers/javascript/packages/anoncreds-nodejs/tests/api.test.ts
```diff
@@ -540,3 +544,326 @@ test('create and verify presentation passing only JSON objects as parameters', (
...
+      schemaId: 'mock:uri',
+      credentialDefinitionId: 'mock:uri',
+      keyCorrectnessProof
+    })
+
+    const linkSecret = LinkSecret.create()
+    const linkSecretId = 'link secret id'
+
+    const { credentialRequestMetadata, credentialRequest } = W3CCredentialRequest.create({
```
TimoGlastra, line 620: Same here

```suggestion
    const { credentialRequestMetadata, credentialRequest } = CredentialRequest.create({
```

wrappers/javascript/packages/anoncreds-nodejs/tests/api.test.ts
```diff
@@ -540,3 +544,326 @@ test('create and verify presentation passing only JSON objects as parameters', (
...
+      entropy: 'entropy',
+      credentialDefinition,
+      linkSecret,
+      linkSecretId,
+      credentialOffer
+    })
+
+    const credential = W3CCredential.create({
```
TimoGlastra, line 628:
```suggestion
    const credential = W3cCredential.create({
```

wrappers/javascript/packages/anoncreds-react-native/src/ReactNativeAnoncreds.ts
```diff
@@ -392,4 +392,192 @@ export class ReactNativeAnoncreds implements Anoncreds {
...
+    const attributeNames = Object.keys(options.attributeRawValues)
+    const attributeRawValues = Object.values(options.attributeRawValues)
+
+    const credential = this.handleError(
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment, @typescript-eslint/prefer-ts-expect-error
+      // @ts-ignore
+      this.anoncreds.createW3CCredential({
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment, @typescript-eslint/prefer-ts-expect-error
+        // @ts-ignore
+        ...serializeArguments(options),
```
TimoGlastra, line 414: Would it be possible to fix these @ts-ignores? 

It will disable any type checking for this line, and can result in type issues down the line

wrappers/javascript/packages/anoncreds-shared/src/api/Credential.ts
```diff
@@ -140,4 +148,32 @@ export class Credential extends AnoncredsObject {
...
+      )
+    } finally {
+      objectHandles.forEach((handle) => {
+        handle.clear()
+      })
+    }
+    return credential
+  }
+
+  public static fromW3C(options: CredentialFromW3COptions) {
```
TimoGlastra, line 176: Won't comment again, but would like to use `w3c` or `W3c` (depending on whether it's the start of the variable /method name), but not use `W3C` (all caps)

wrappers/javascript/packages/anoncreds-shared/src/api/W3CCredential.ts
```diff
@@ -0,0 +1,231 @@
...
+    })
+
+    this.handle.clear()
+    this.handle = credential
+  }
+
+  public addType(type_: string) {
+    const credential = anoncreds.w3cCredentialAddType({
+      objectHandle: this.handle,
+      type_
```
TimoGlastra, line 225: In JS it's fine to use `type` as a variable name I think. Is this taken over from the Python wrapper?

Would prefer to just use `type`, or otherwise if you want to avoid using `type`, use something like `credentialType`

TimoGlastra (changes requested):
Mostly looked at the JS wrapper ,as I have more experience with it. 

I think overall looks good, but I think we can reduce the number of method exposed in the wrapper, as well as have some more consistency in using w3c in names

Artemkaaas:
@TimoGlastra I processed and resolved comments connected to using `W3c` instead `W3C`. 
I will handle the rest of comments (relating to credential-offer/request duplication and credential methods for non-anoncreds proof) after discuss at the WG call. Unfortunately, I may miss today's call due to some personal errands.

wrappers/javascript/packages/anoncreds-nodejs/src/NodeJSAnoncreds.ts
Artemkaaas: I do not understand your point. `w3cCredentialGetAttribute` method calls `anoncreds_w3c_credential_get_attribute` external Rust function which obtain W3C credential object from handle and returns value for requesting attribute.

TimoGlastra: Yeah I was thinking more of that whether the object is an AnonCreds representation or a W3C representation doesn't really change the input or output of this method, so from an user POV it could be the same method right?

wrappers/javascript/packages/anoncreds-nodejs/src/NodeJSAnoncreds.ts
Artemkaaas: I added helper methods doing conversion of parameters in order to reduce code duplication. 
But left w3c* methods. 
I think methods here should be mapped one to one with external C API. Do not see a sense to combine w3c and legacy methods.

TimoGlastra:
@Artemkaaas once #291 is merged, I'll take another look at this PR, but there's currently too much different changes

swcurran:
The plan (per requests from @TimoGlastra and @andrewwhitehead ) is to merge #273 first (once the test failure is fixed), so this is easier to check. Tests are passing here, so hopefully this one will be easy to approve once #273 is merged.

FYI @Artemkaaas 

Thanks

swcurran:
@TimoGlastra , @andrewwhitehead -- can you please do a review of this PR. The other two have been merged, so this should be easier to review.  Thanks.

docs/design/w3c/w3c-representation.md
```diff
@@ -343,297 +297,3 @@ w3c_presentation_request = Verifier.w3c_create_presentation_request()
...
-    )
-    if !credential {
-        error
-    }
-    credential.checkRestrictions(requested.restrictions)
-    assert(credential.credentialSubject[requested.name])
-}
-```
-
-### Examples
```
andrewwhitehead, line 546: Do we not want to keep the examples?

wrappers/javascript/packages/anoncreds-shared/src/api/index.ts
```diff
@@ -15,3 +15,6 @@ export * from './RevocationRegistry'
 export * from './RevocationRegistryDefinition'
 export * from './RevocationRegistryDefinitionPrivate'
 export * from './RevocationStatusList'
+export * from './W3cCredential'
+export * from './CredentialOffer'
```
andrewwhitehead, line 19: CredentialOffer is already exported

wrappers/python/anoncreds/types.py
```diff
@@ -342,6 +343,166 @@ def rev_reg_index(self) -> Optional[int]:
...
+    def from_legacy(
+        cls,
+        cred: "Credential",
+        cred_def: Union[str, CredentialDefinition],
+        w3c_version: Optional[str] = None
+    ) -> "W3cCredential":
+        return cred.to_w3c(cred_def, w3c_version)
+
+    def _get_proof_details(self) -> bindings.ObjectHandle:
+        if self._proof_details == None:
```
andrewwhitehead, line 448: Comparisons to None should use 'is'

andrewwhitehead:
Looks good, with a couple tiny notes. We'll want to add the w3c tests to the CI.

TimoGlastra (approved):
LGTM! I already played with the api a bit for creating the test vectors and it made sense to me

Only thing is maybe that there's quite some duplication between the w3c and non w3c code in the JS wrapper, but we can look at that in the future maybe. Would ne good to have this merged
And reelased!

TimoGlastra:
Also one more question/
: did you test the react native code already? Or should we do some
Testing in react native?

Artemkaaas:
> Also one more question/ : did you test the react native code already? Or should we do some Testing in react native?

I only checked build and compile scripts for wrapper, but not tested integration into a react-native application.

TimoGlastra:
> > Also one more question/ : did you test the react native code already? Or should we do some Testing in react native?
> 
> I only checked build and compile scripts for wrapper, but not tested integration into a react-native application.

We can do that once this is merged and released as dev 👍

PR #276, opened 2023-11-22
AnonCreds Credentials using the W3C Standard - proof value encoding

Artemkaaas:
Use message pack for encoding proof values of w3c credentials and presentation.
Results of proof value string length on example of credentials signature:
* base64(JSON) previously used - 3821 
* base64(**MessagePack**) will be used - 2541
* base64(CBOR) - 3041


TimoGlastra (approved):

src/data_types/nonce.rs
```diff
@@ -187,9 +196,22 @@ impl<'a> Deserialize<'a> for Nonce {
...
             }
+
+            fn visit_seq<E>(self, mut seq: E) -> Result<Self::Value, E::Error>
+            where
+                E: SeqAccess<'a>,
+            {
+                let mut vec = Vec::new();
+
+                while let Ok(Some(Value::Number(elem))) = seq.next_element() {
+                    vec.push(elem.as_u64().unwrap() as u8);
```
andrewwhitehead, line 207: This should likely be more tolerant of bad inputs, not panicking via `unwrap`. It should also return an error if a non-Number element is found.

Artemkaaas: fixed

andrewwhitehead (approved):

andrewwhitehead:
Tests are failing because a Rust dependency now requires 1.66 or higher. We can either update the CI rust version or add a lockfile in another PR.

swcurran:
Seems like updating would be the prudent thing?  What is typically done in this situation?  How much risk is bumping the rust version?  As I recall — the. #295  is failing one test (Android) because it needs 1.70.  Should we bump to that?

andrewwhitehead:
There's no real risk in bumping the Rust version used, it just means that we can't guarantee compatibility with older versions. I think the Android failure is different (one of cross's dependencies), and partly fixed by this PR, but it would eventually need a lockfile to maintain its platform support.

swcurran:
OK — just trying to figure out how we get to a decision/action on this. Who makes the call?

TimoGlastra:
@berendsliedrecht just checking if this can impose a problem for the supported versions of the RN wrapper if the version is bumped?

swcurran:
@Artemkaaas --- sorry to be a pain. Looks like we have one unit test failing. Could you please take a look?

> test data_types::w3c::proof::tests::encoded_object_encode_decode_works ... FAILED

Fails on both Ubuntu and Windows.

PR #277, opened 2023-11-22
Address SNYK warnings

ryjones:


berendsliedrecht (approved):

swcurran:
@berendsliedrecht — what do you think of the test failures? Are they relevant to the change, or gremlins.

Thanks, @ryjones — I should have done that, and will next time.  I just thought there might be some sort of setting that enabled dependabot-type, DCO-loving PRs to be produced from Snyk :-).

berendsliedrecht:
@swcurran let me have a look. I think we generate some number non-deterministically which made the test fail this one time (assumption based on the error below).

`thread 'main' panicked at 'Unable to create or update the revocation state: Error { kind: InvalidState, cause: None, message: Some("Revocation index is outside of valid range") }', src\services\prover.rs:64:49`

Will hopefully provide a fix asap.

PR #284, opened 2023-12-06
AnonCreds Credentials using the W3C Standard - drop encoding from credential schema

Artemkaaas:
Drop `encoding` property from the credential schema as a specific one will be set at the anoncreds specification


Artemkaaas:
Closed in favour of: https://github.com/hyperledger/anoncreds-rs/pull/291

PR #285, opened 2023-12-06
Manual update of Synk PRs that are not DCOd

swcurran:
Address PRs:

- #283 
- #279 
- #278

### DIFF
--- a/wrappers/javascript/packages/anoncreds-react-native-example/package.json
+++ b/wrappers/javascript/packages/anoncreds-react-native-example/package.json
@@ -11,10 +11,10 @@
   "dependencies": {
     "@hyperledger/anoncreds-react-native": "workspace:*",
     "expo": "~49.0.10",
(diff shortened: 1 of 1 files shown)


swcurran:
@berendsliedrecht @genaris — I’m at the limit of my (very limited!!) knowledge on this.  I’ve updated the two package.json files (anoncreds-react-native, anoncreds-react-native-example) folders, and update the lock file via a “pnpm install”.  Now the GHA is failing on another error and (“prettier” — list differences) — I don’t know what to do about that.

Can you help?

berendsliedrecht:
Should be fixed now. We should actually not push the pnpm lock file, but that might cause some other CI issues, so this'll do for now.

swcurran:
OK.  The file was already there and I updated it. I’ll leave it to you to decide if anything else is needed.

Thanks!!!

PR #291, opened 2023-12-20
Reworked w3c credentials and presentations to use DataIntegrityProof  instead of custom AnonCreds context

Artemkaaas:
Update W3C Credential to match the design: https://github.com/hyperledger/anoncreds-spec/issues/192

I also included changes from the following PRs as their changes significantly intersected with the current one: 
* Drop encoding - https://github.com/hyperledger/anoncreds-rs/pull/284
* Drop mapping - https://github.com/hyperledger/anoncreds-rs/pull/274

Regarding vc format `1.1`/`2.0:` I added an option parameter `version` to `anoncreds_process_w3c_credential`, `anoncreds_credential_to_w3c`, and `anoncreds_create_w3c_presentation` functions defining what vc format to return.


design.md
TimoGlastra, line 1: This file seems empty?

docs/design/w3c/w3c-representation.md
```diff
@@ -280,20 +200,21 @@ pub extern "C" fn anoncreds_process_w3c_credential(
...
+/// Get credential signature information required for proof building and verification
+/// This information is aggregated from `anoncredsvc-2023` and `anoncredspresvc-2023` proofs.
+/// It's needed for Holder and Verifier for public entities resolving
+///     {`schema_id`, `cred_def_id`, `rev_reg_id`, `rev_reg_index`, `timestamp`}
 ///
 /// # Params
 /// handle:                object handle pointing to the credential (in W3 form)
-/// name:                  name of attribute to retrieve
-/// result_p:              reference that will contain value of request credential attribute
+/// result_p:              reference that will contain credential information
```
TimoGlastra, line 210:
```suggestion
/// cred_proof_info_p:              reference that will contain credential information
```

src/data_types/w3c/credential.rs
```diff
@@ -73,356 +55,120 @@ pub struct CredentialSubject {
...
-            type_: CredentialSchemaType::AnonCredsDefinition,
-            definition,
-            schema,
-            encoding,
+        issuer: IssuerId,
+        attributes: CredentialAttributes,
+        proof: DataIntegrityProof,
+        version: Option<&VerifiableCredentialSpecVersion>,
+    ) -> Self {
+        let time = Utc::now();
```
TimoGlastra, line 75: validFrom is not required in v2, so I think we should leave it out in this case. Adding it for the v1 one version is just a 'hack' I think to make it compliant with the data model

src/data_types/w3c/credential.rs
```diff
@@ -73,356 +55,120 @@ pub struct CredentialSubject {
...
-            proof: OneOrMany::Many(Vec::new()),
-            ..Default::default()
+            context: credential.context.clone(),
+            type_: credential.type_.clone(),
+            issuer: credential.issuer.clone(),
+            id: credential.id.clone(),
+            issuance_date: credential.issuance_date,
+            valid_from: credential.valid_from,
+            credential_subject: CredentialSubject {
+                id: credential.credential_subject.id.clone(),
```
TimoGlastra, line 109: Is this method used to create a derived credential? I don't think we always want to include the `id` property in a derived credential? Or maybe I'm misunderstanding this code incorrectly.

src/data_types/w3c/credential.rs
```diff
@@ -73,356 +55,120 @@ pub struct CredentialSubject {
...
+                if self.issuance_date.is_none() {
+                    return Err(err_msg!(
+                        "V1.1 Credential must include `issuanceDate` property"
+                    ));
+                }
+            }
+            VerifiableCredentialSpecVersion::V2_0 => {
+                if self.valid_from.is_none() {
+                    return Err(err_msg!(
+                        "V1.1 Credential must include `validFrom` property"
```
TimoGlastra, line 166: `validFrom` is not required. https://www.w3.org/TR/vc-data-model-2.0/

```suggestion
                        "V2 Credential must include `validFrom` property"
```

TimoGlastra (approved):
I tried to go over it, but there's a lot of changed code, so it's a bit hard to look really in-depth whether are changes are correct. But overall looks good to me!

src/data_types/w3c/credential.rs
Artemkaaas: Correct.   This method is used to construct a credential which will be put into a presentation.

TimoGlastra: Then we should remove the `id` property I think, otherwise it reveals a correlatable identifier. The `id` property should only be revealed when it's revealed in the crednetial.

TimoGlastra:
I think one thing that is missing to make it compliant with the DI spec is that the proofValue MUST be multibase encoded and have a header:

> A string value that contains the base-encoded binary data necessary to verify the digital proof using the verificationMethod specified. The contents of the value MUST be expressed with a header and encoding as described in Section [2.4 Multibase](https://www.w3.org/TR/vc-data-integrity/#multibase-0). The contents of this value are determined by a specific cryptosuite and set to the proof value generated by the [Add Proof Algorithm](https://www.w3.org/TR/vc-data-integrity/#add-proof) for that cryptosuite. Alternative properties with different encodings specified by the cryptosuite MAY be used, instead of this property, to encode the data necessary to verify the digital proof.

I don't know if there's a multibase format we can use that won't add too much overhead to the final proofValue? It must decode to the binary data needed to verify the proof. I don't know if the output of messagepack is binary? If so, we could do a base64 pass on it and use that as the proofValue? 

Here's a list of all multibase types: https://github.com/multiformats/multibase/blob/master/multibase.csv

Artemkaaas:
@TimoGlastra I already applied multibase encoding for the proof value with `u` prefix (base64 no padding) here: 

https://github.com/hyperledger/anoncreds-rs/pull/291/files#diff-0423f5bfbbb105eb8c19ba64df62479a284b630ed112415e211756ac7b2d38d1

```
const BASE_HEADER: char = 'u';

pub trait EncodedObject {
    fn encode(&self) -> String
    where
        Self: Serialize,
    {
        let json = json!(self).to_string();
        let serialized = base64::encode(json);
        format!("{}{}", BASE_HEADER, serialized)
    }

    fn decode(string: &str) -> Result<Self>
    where
        Self: DeserializeOwned,
    {
        match string.chars().next() {
            Some(BASE_HEADER) => {
                // ok
            }
            value => return Err(err_msg!("Unexpected multibase base header {:?}", value)),
        }
        let decoded = base64::decode(&string[1..])?;
        let obj: Self = serde_json::from_slice(&decoded)?;
        Ok(obj)
    }
}
```

TimoGlastra:
Oh woops, sorry completely missed that.

Nice 😬

swcurran:
@Artemkaaas — there is a conflict in the PR now. Could you please resolve and we’ll merge this.  Sorry for the delay over the break.

PR #298, opened 2024-01-10
Update Rust version used in CI

andrewwhitehead:
This adjusts the CI to use the stable Rust release from 6 months ago (currently 1.70.0) so that maybe it doesn't need to be updated manually.

The installation of cross is adjusted to use its lockfile, in line with indy-vdr and askar.


andrewwhitehead:
Unlike the other repos, this one does not appear to override the Rust version in the Android build for platform version compatibility. That could be added in another PR, but would need a lockfile to be generated with that Rust version.

berendsliedrecht (approved):

PR #302, opened 2024-01-15
W3C: Reworked predicates representation

Artemkaaas:
Reworked predicates representation in presented credential to use boolean `true` value as the result of a predicate instead of custom object.


src/services/w3c/verifier.rs
```diff
@@ -135,154 +136,106 @@ fn check_credential_restrictions(
...
                 schemas,
                 cred_defs,
-                predicate.non_revoked.as_ref(),
-                nonrevoke_interval_override,
                 proof,
-            )
-            .is_ok();
-
-        if valid_credential {
+            )?;
```
TimoGlastra, line 238: Sorry for my ignorance, I'm probably missing something here, but I still not quite understand why the change of predicate from an object to boolean has made it required to add the mapping back into the proof. We can still infer this based on the submitted proof itself and the proof request right? The proof contains proof for predicate X, while the proof request also contains a requirement for predicate x > 50 for example. 

Won't it be possible to match these against each other? Why was it possible with the object structure?

Artemkaaas: Due to the corner case, Stephen mentioned here: https://github.com/hyperledger/anoncreds-spec/issues/194#issuecomment-1876070195

The same name of attribute was requested twice, but different credentials were used for the proof .
Here is how the request will look like:
```
{
    "name":"test",
     "version": "1.0",
     "nonce": "1234556"
     "requested_predicates”: { 
     	 “predicate_1”: {
		“name”: “age”,
 		“p_type”: “>=”,
		“p_value”: 18
         },
     	 “predicate_2”: {
		“name”: “age”,
 		“p_type”: “>”,
		“p_value”: 20
         }
}
```     

`anoncreds-clsignatures` requires passing of predicate name, type, value for doing verification. If we pass different value, verification will fail.


 
Here is a proof verification steps using `anoncreds-clsignatures` methods: 
```
//  taken from proof request
//   pass predicate name/type/value   
let mut sub_proof_request_builder = Verifier::new_sub_proof_request_builder().unwrap();
sub_proof_request_builder.add_predicate("age", "GE", 18).unwrap();
let sub_proof_request = sub_proof_request_builder.finalize().unwrap();

// taken from presentation
let mut proof_verifier = Verifier::new_proof_verifier().unwrap();
proof_verifier.add_sub_proof_request(&sub_proof_request, &credential_schema, &non_credential_schema, &credential_pub_key,  None, None).unwrap();
proof_verifier.verify(&proof, &proof_request_nonce).unwrap()
```

PR #303, opened 2024-01-15
W3C: Reworked predicates representation - no mapping

Artemkaaas:
@TimoGlastra @andrewwhitehead  @swcurran 

This is an alternative implementation for w3c predicate representation as boolean values to using explicit mapping like was done in PR: https://github.com/hyperledger/anoncreds-rs/pull/302 

I feel like it should work and we can close another PR.

`anoncreds-clsignatures` requires passing of predicate name, type, value for doing verification of each cl sub-proof. 
If we pass a different predicate value, verification will fail.
It can cause an issue when a request contains multiple predicates with the same attribute name. Due to this, we used an explicit mapping.  

 In proper way, this information (predicate name /type/value) should be taken from the proof request.
 But as we do verification for each request attribute/predicate that presentation contains a verifiable credential providing requested data, we can just take `attributes/predicates` from inside of sub proof itself at the sub proof verification step . 

The function to verify requested predicate check the following:
1. w3c credential contains a predicate attribute in credential subject
2. cl `sub_proof` contains sub_proof for the requested predicate (same name, value, type)
3. credential matches to restrictions

This PR: depends on this change in CL AnonCreds library: https://github.com/hyperledger/anoncreds-clsignatures-rs/pull/38


Cargo.toml
```diff
@@ -27,7 +27,7 @@ vendored = ["anoncreds-clsignatures/openssl_vendored"]
 w3c = ["base64", "chrono", "rmp-serde"]
 
 [dependencies]
-anoncreds-clsignatures = "0.3.0"
+anoncreds-clsignatures = { git = "https://github.com/DSRCorporation/anoncreds-clsignatures-rs.git", branch = "feat/expose-proof-predicates" }
```
TimoGlastra, line 30: 0.3.1 has been released

```suggestion
anoncreds-clsignatures = "0.3.1"
```

TimoGlastra (approved):

TimoGlastra:
Merging for now so we can create a pre-release. @andrewwhitehead if you have any comments, we can addreess them in a follow up pr

PR #317, opened 2024-02-01
Use a single cryptosuite identifier and tag proofValue with subtype

### LINKED ISSUE #315: Data Integrity Proof Cryptosuite naming
I know this has been discussed before, but I would like to reiterate the current use of Cryptosuite properties when working with w3c credentials and presentations.

Currently, we use `anoncreds-vc` for credentials, `anoncredspres-vc` for derived credentials, and `anoncredspresvp-2023` for presentations.

There are a few issues with this approach.
1. Having multiple different identifiers here naturally introduces some complexity.
    E.g., checks and mappings for these identifiers.
2. It is not really in line with what the Data Integrity spec defines as a crypto suite,
    an identifier for a `set of specific cryptographic primitives,` as we don't change the cryptographic primitives.
3. Also, other crypto suites supporting selective disclosure don't follow this approach, e.g., `https://www.w3.org/TR/vc-di-bbs/`.

During a working group call, it was suggested to use only a single crypto suite value,
e.g. by encoding the required metadata in the proofValue. I think it would make sense to move in this direction.

andrewwhitehead:
Fixes #315

This changes the cryptosuite value to "anoncreds-2023".

The proofValue of the DataIntegrityProof is encoded as a 2-element list with the subtype as the first element: 1=credential signature, 2=credential presentation, 3=presentation proof (aggregate proof). This makes for a small and compatible representation, although other options could be considered, such as a map (`{"CredentialSignature": {...}}`).


TimoGlastra (approved):
Nice, thanks @andrewwhitehead

PR #320, opened 2024-02-05
refactor: only issuer id for legacy->w3c

TimoGlastra:
We were running into some issues when needing to transform between legacy -> w3c anoncreds as the method requires the cred_def to be passed. It seems the cred_def is only required for the issuer_id, and this PR updates it to only require a string issuer_id rather than the whole cred def. 

In our case we know the issuer_id, but we'd have to resolve the cred_def from the ledger

### DIFF
--- a/src/ffi/w3c/credential.rs
+++ b/src/ffi/w3c/credential.rs
@@ -1,3 +1,4 @@
+use crate::data_types::issuer_id::IssuerId;
 use crate::data_types::w3c::VerifiableCredentialSpecVersion;
 use ffi_support::{rust_string_to_c, FfiStr};
 use std::ffi::c_char;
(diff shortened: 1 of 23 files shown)


src/ffi/w3c/credential.rs
```diff
@@ -148,8 +149,8 @@ pub extern "C" fn anoncreds_credential_to_w3c(
             None => None,
         };
 
-        let w3c_credential =
-            credential_to_w3c(credential, cred_def.load()?.cast_ref()?, w3c_version)?;
+        let issuer_id = IssuerId::new(issuer_id).expect("Invalid issuer ID");
```
TimoGlastra, line 152: I'm not a rust expert. Is this validated and handled correctly?

berendsliedrecht: If new returns an error it will panic. You can use `?` instead to not panic but pass along the error back over ffi.

TimoGlastra: Thanks! Will update

PR #331, opened 2024-04-05
fix(js): update ffi-napi

genaris:
Updated `ffi-napi` dependency on JS wrapper in order to fix the incompatibility issue found with newer node.js versions. See https://github.com/nodejs/node/issues/52240#issuecomment-2040598178.

### DIFF
--- a/wrappers/javascript/packages/anoncreds-nodejs/package.json
+++ b/wrappers/javascript/packages/anoncreds-nodejs/package.json
@@ -27,7 +27,7 @@
     "install": "node scripts/install.js"
   },
   "dependencies": {
-    "@2060.io/ffi-napi": "4.0.8",
+    "@2060.io/ffi-napi": "4.0.9",
     "@2060.io/ref-napi": "3.0.6",
     "@hyperledger/anoncreds-shared": "workspace:*",
     "@mapbox/node-pre-gyp": "^1.0.11",
--- a/wrappers/javascript/packages/anoncreds-nodejs/package.json
+++ b/wrappers/javascript/packages/anoncreds-nodejs/package.json
@@ -27,8 +27,8 @@
     "install": "node scripts/install.js"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


swcurran:
Any idea on the failed test?  Looks unrelated to what you did, but...

genaris:
> Any idea on the failed test? Looks unrelated to what you did, but...

It seems that it already failed on the last commit, around 45 days ago: https://github.com/hyperledger/anoncreds-rs/actions/runs/8031489343/job/21939863984

If we look at the [latest successful AnonCreds Lint action](https://github.com/hyperledger/anoncreds-rs/actions/runs/7965323269/job/21744609961), we'll find that it was installing a different rustc version:

- Good one: 
>  rustup toolchain install 1.71 --component clippy --component rustfmt --profile minimal --no-self-update
 
- Bad one:
>  rustup toolchain install 1.73 --component clippy --component rustfmt --profile minimal --no-self-update

This happens because in the CI we are setting `RUST_VERSION="stable 6 months ago"`, which of course is variable unless we find a way to stop the time (which would be great by the way!).

@andrewwhitehead @berendsliedrecht do you think we can stick to 1.71? Or do you find better to see why this format check is failing with 1.73 and fix it?

berendsliedrecht (approved):

wrappers/javascript/packages/anoncreds-nodejs/package.json
```diff
@@ -27,7 +27,7 @@
     "install": "node scripts/install.js"
   },
   "dependencies": {
-    "@2060.io/ffi-napi": "4.0.8",
+    "@2060.io/ffi-napi": "4.0.9",
```
TimoGlastra, line 30: Maybe we should use caret versions so we can publish fixes to the ffi-napi in the future without having to make new releases for the wrappers

TimoGlastra (approved):

PR #341, opened 2024-05-31
feat: added support for github actions

wadeking98:
added github action to push alpha releases 
This doesn't affect the normal release cycle, only alpha releases after something gets merged


wrappers/javascript/packages/anoncreds-react-native/package.json
cvarjao, line 1: We also need to remove `@mapbox/node-pre-gyp`

TimoGlastra (changes requested):
Same comments as Indy VDR PR. Let's first figure that one out and then we can apply the same changes here

PR #342, opened 2024-06-26
chore(js): update version to 0.2.3

TimoGlastra:
Update version so we can release https://github.com/hyperledger/anoncreds-rs/pull/338

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -207,17 +207,17 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Set up Node.JS 18.x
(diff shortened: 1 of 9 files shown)


swcurran:
Hey @TimoGlastra — what should happen with the error in the wrapper testing?  Fix something?  Weird looking error messages, but presumably all stemming from the same thing?

berendsliedrecht (approved):
Think you have to run prettier :).

TimoGlastra:
Running prettier /lint locally does not give the same error

berendsliedrecht:
> Running prettier /lint locally does not give the same error

Different versions? Not sure what else it could be.

genaris (approved):
Working now I've upgraded pnpm version to 9 in CI

PR #355, opened 2025-03-10
Restored get_attribute for creddef and updated python wrapper

### LINKED ISSUE #354: restore anoncreds_credential_definition_get_attribute
In the commit e5099af809747e0bab46113dd9c5fa67c1854748 we seem to have unintentionally removed `anoncreds_credential_definition_get_attribute`

Gavinok:
This pull request resolves #354

I have tested this against the python wrapper and updated it to support the current values of the CredentialDefinition struct


berendsliedrecht:
rebasing with main should fix the build issues

Gavinok:
@berendsliedrecht rebased. Thanks for letting me know

wrappers/python/anoncreds/types.py
```diff
@@ -40,25 +40,46 @@ def load(cls, value: Union[dict, str, bytes, memoryview]) -> "CredentialDefiniti
         )
 
     @property
-    def id(self) -> str:
+    def schema_id(self) -> str:
```
andrewwhitehead, line 43: This seems to drop the "id" property accessor?

Gavinok: To my knowledge there isn't actually an id field in CredentialDefinition.
https://github.com/Gavinok/anoncreds-rs/blob/main/src/data_types/cred_def.rs#L39-L48

Feel free to correct me if I am missing something

Issue #365, opened 2025-09-12
Issue with Python Audit wheel


TimoGlastra:
There seems to  be an error in CI related to the Python wrapper.

> Run auditwheel show wrappers/python/dist/* | tee auditwheel.log
>
> anoncreds-0.2.1-py3-none-manylinux2014_aarch64.whl is consistent with
> the following platform tag: "manylinux_2_28_x86_64".
>
> The wheel references external versioned symbols in these
> system-provided shared libraries: libm.so.6 with versions
> {'GLIBC_2.27'}, libdl.so.2 with versions {'GLIBC_2.17'},
> libpthread.so.0 with versions {'GLIBC_2.17'}, libc.so.6 with versions
> {'GLIBC_2.18', 'GLIBC_2.25', 'GLIBC_2.17', 'GLIBC_2.28'},
> libgcc_s.so.1 with versions {'GCC_4.2.0', 'GCC_3.0', 'GCC_3.3'}
>
> This constrains the platform tag to "manylinux_2_28_x86_64". In order
> to achieve a more compatible tag, you would need to recompile a new
> wheel from source on a system with earlier versions of these
> libraries, such as a recent manylinux image.
> Error: Process completed with exit code 1.

https://github.com/hyperledger/anoncreds-rs/actions/runs/17647852499/job/50151538648

@andrewwhitehead any idea how to resolve this?

TimoGlastra:
Hmm it seems the CI uses `anoncreds-ubuntu-latest` custom runner. So we can't downgrade. I have no idea where this image / runner is coming from.

@ryjones do you know where this runner is configured and managed?

ryjones:
We really need to migrate your repos to https://github.com/hyperledger-anoncreds to make things easier

genaris:
> We really need to migrate your repos to https://github.com/hyperledger-anoncreds to make things easier

Is it needed to migrate all the repos at once, or can we just move this one for the moment? We are in a hurry right now, since we need to get this thing fixed (Google imposed a deadline for november, and we need to give our users at least some weeks to prepare to upgrade their apps).

What do you suggest to do to make it work?

TimoGlastra:
[...]

PR #367, opened 2025-10-01
chore: v0.2.2

genaris:
Just a version bump prior to triggering a new release. Although we still have the issue of creating Python wrapper (#365), I think we can create a new release to publish mobile binaries required by JavaScript wrapper.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "anoncreds"
-version = "0.2.1"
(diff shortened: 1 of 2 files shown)


swcurran (approved):

TimoGlastra:
> I think we can create a new release to publish mobile binaries required by JavaScript wrapper.

How do we bypass the ci issues? Disable them for now?

